PR #21, opened 2017-06-07
Fix to avoid excess local variable updates

slacrherbst:


slacrherbst:
Ok. I will do so with a separate pull request.

PR #23, opened 2017-06-07
added device resets

slacrherbst:
This is to replace setResetFunc now that we are properly subclassing devices. The user should now just override the softReset, hardReset and countReset functions in a device. We will eventually get rid of setResetFunc.


slacrherbst:
The soft, hard and count reset commands are exposed at the root as commands.

bengineerd:
Ahh, I see.

PR #30, opened 2017-06-09
Add ability to override block processing at the device level.

slacrherbst:
I believe this is ready to go if you agree with the new structure.


slacrherbst:
Merged this pull request into #34

PR #33, opened 2017-06-09
Add support for sparse bit fields

slacrherbst:


slacrherbst:
The user has to be carefull to ensure that any other variables that touch the register space in 0x40 - 0x4C all use offset = 0x40 with the appropriate bitOffset. I need to work on making addBlocks smarter about this. I will think about that tonight and tomorrow.

slacrherbst:
I don't fully understand the question about block.py line 279. For the second question I am not sure what the field masks would look like. The field would still be generated bit by bit, correct?

slacrherbst:
Ah I think you mean line 297. Yes you are correct.

slacrherbst:
Merged this pull request into #34

PR #34, opened 2017-06-12
Support for sparse bit fields & ability to override block processing at the device level

slacrherbst:
Combination of pull requests #33 and #30


bengineerd:
Is there a requirement that the lists passed to offset, bitOffset and bitSize be in some kind of increasing order? The stuff I just added won't care, but maybe your block mapping code does. I could sort them for you if needed.

slacrherbst:
No. As long as they are associated with each other the order does not matter. Since you already normalize everything to min(offset) the blockCreation is fine. I also don't care about the order of the variable adds since I sort the variables by offset,size.

bengineerd:
I think everything is working for my Coulter project. If it all works for you too we should go ahead and merge.

bengineerd:
68 commits!
I think we're ready to merge.

PR #35, opened 2017-06-13
Add range and wildcard to yaml load

slacrherbst:
An index array (Test[10:20]) can be accessed in the following ways:

Test = Applies to all
Test[*] Applies to all
Test[10:12] Applies to 10 and 11
Test[:12] Applies to 10 and 11
Test[19:] Applies to 19 and 20


bengineerd:
Maybe we use [:] to apply to all instead of [*]? Is that more consistent?

I've always found python's slice syntax confusing on the upper bound. It's inclusive on the lower but exclusive on the upper. You follow the python syntax, which I think we want to do, but I bet a lot of people will try Test[10:11] to set both 10 and 11.

slacrherbst:
When I first started I was parsing the fields myself. Instead I chose to use the python eval statement, with the exception that I added the ability to include the '*' wildcard. I would be happy to remove the wildcard.

PR #36, opened 2017-06-13
Moved req transaction in lock scope. start timer after return

slacrherbst:
This breaks my general rule of only holding one lock at a time but I think it is necessary (_reqTransaction will grab a lock). The lock on block should not be released until the transaction gets queued. Otherwise if the queue blocks other processes can start a transaction on the block in parallel. Also the timer should not start until the transaction is queued.


slacrherbst:
Deferring.

PR #44, opened 2017-06-18
Added helper functions for making simple commands

ruck314:


slacrherbst:
I think the one function has an error. 

For 's in sets
   Cmd.set (i)

Should that be 

Cmd.set (s)?

PR #47, opened 2017-06-20
Fix Polling bugs - ESROGUE-54

bengineerd:


slacrherbst:
How was var update being called before root was attached? May be another bug which is hidden by this fix. Maybe we need to add a print if var is called before root is attached. Or we should just make self._root = None by default and just chase the errors? For now this is fine.

bengineerd:
It gets called through the poller.

_rootAttached gets called on a Variable, and it's Block gets added to the poller.
Then the poller fires a read on that block right away, before the root has been attached to other Variables that share the Block.
Then _updated() gets called on all the Variables of the Block that the poller read.

slacrherbst:
Should we add a call to root which starts polling after setup is done?

On June 21, 2017 9:02:30 AM GMT+10:00, Benjamin Reese <notifications@github.com> wrote:
>It gets called through the poller.
>
>_rootAttached gets called on a Variable, and it's Block gets added to
>the poller.
>Then the poller fires a read on that block right away, before the root
>has been attached to other Variables that share the Block.
>Then _updated() gets called on all the Variables of the Block that the
>poller read.
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/slaclab/rogue/pull/47#issuecomment-309915915

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

bengineerd:
But how does it know that the setup is done?

Maybe we need a pyrogue.createApp(root) function, that can make a specified set of calls to root once all of the Node creation and tree building is done.

slacrherbst:
The user calls it.

For example you can't start epics or Pyro until setup is done.

So the user calls enable polling as the last line in the init.

On June 21, 2017 9:33:17 AM GMT+10:00, Benjamin Reese <notifications@github.com> wrote:
>But how does it know that the setup is done?
>
>Maybe we need a pyrogue.createApp(root) function, that can make a
>specified set of calls to root once all of the Node creation and tree
>building is done.
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/slaclab/rogue/pull/47#issuecomment-309920768

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #51, opened 2017-06-30
Create start() method for tree and cleanup rootAttached

slacrherbst:


slacrherbst:
I posted a change which limits the deprecation dump to 50.

PR #57, opened 2017-07-07
Remove rawblock and fix bugs in last pull request

slacrherbst:
rawBlock has been removed. It has been replaced with rawWrite and rawRead calls in the Device class.

So instead of creating the rawBlock object and calling write and read on it, simply call self.rawWrite and self.rawRead with the same api. Performance has improved a bit. ~12Khz at 10Gbps.

This also fixes a bug I found in the last pull request which was causing some memory transactions to have undefined errors. waitTransaction() was not truly waiting for the transaction to complete. 

recompile is required.

### DIFF
--- a/include/rogue/interfaces/memory/Hub.h
+++ b/include/rogue/interfaces/memory/Hub.h
@@ -55,9 +55,6 @@ namespace rogue {
                //! Destroy a hub
                ~Hub();
 
-               //! Get address
-               uint64_t getAddress();
-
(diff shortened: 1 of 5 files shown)


slacrherbst:
Ok. I moved rawWrite and rawRead to private. Not sure about the base parameter. My assumption is that these two calls are used for very particular cases where some level of conversion is already needed in the Device. We could add at some point if we feel it is needed.

PR #60, opened 2017-07-10
Added catch for using stride in wrong method

slacrherbst:


bengineerd:
Invalidated by another PR.

PR #63, opened 2017-07-11
ESROGUE-60 - Efficient RAM device

bengineerd:
https://jira.slac.stanford.edu/browse/ESROGUE-60

Any Device now has raw access to its address space with rawRead() and rawWrite().
There is also a MemoryDevice class that will facilitate loading large memory configs through YAML.


bengineerd:
So we'll merge this but keep the branch open to further develop MemoryDevice.

PR #66, opened 2017-07-13
Device size

slacrherbst:


slacrherbst:
@ruck314 This will require a size when creating an empty device which will use rawWrite and rawRead. Normally the device size will be determined by its variables. For an empty device you need to pass size so we can do valid range checking to avoid overflows.

PR #77, opened 2017-07-19
ESROGUE-60 and other fixes

bengineerd:
This branch was nominally for ESROGUE-60, but a lot of other stuff ended up in it.

### ESROGUE-60
Created a new MemoryDevice that does all of its memory access via raw reads and writes rather than through Variables. Can be configured via YAML.

### ESROGUE-77
Raw writes and read can now be issued that are larger than the underlying maxAccess size. They will be split up into smaller chunks automatically.

### Low level changes
- Many performance tweaks in interfaces/memory/Master.cpp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@
 # ----------------------------------------------------------------------------
 
 VER_MAJOR := 2
-VER_MINOR := 1
-VER_MAINT := 1
+VER_MINOR := 2
+VER_MAINT := 0
 
 # Variables
 CC       := g++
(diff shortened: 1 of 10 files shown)


bengineerd:
I think we need to have Larry test it with his cryo-det stuff first.

Also, add the Device size check stuff back in.

slacrherbst:
I think we should push it and be ready for an issues that come up. I will be gone next week so best to deal with errors this week. My testing shows it works. The new features may have issues.

slacrherbst:
I would add the size checking after the pull request as a bug fix.

ruck314:
>> I think we need to have Larry test it with his cryo-det stuff first.

I have tested it with Ben.  Seems to work for the cryo-det platform

PR #81, opened 2017-07-19
raw read fix

slacrherbst:


slacrherbst:
Ok. That would at least remove an if statement. I backed it out because the if statement was redundant.

PR #95, opened 2017-08-03
Esrogue 91

jesusvasquez333:
As EPICS uses only signed 32-bit integers, we need to cast the values when the rogue variables use unsigned value instead.

### DIFF
--- a/python/pyrogue/epics.py
+++ b/python/pyrogue/epics.py
@@ -40,17 +40,6 @@ class EpicsCaDriver(pcaspy.Driver):
         self._q.put(entry)
         self.setParam(reason,value)
 
-    def read(self,reason):
-        # Read the value from the parameter list
-        value = self.getParam(reason)
-
-        # EPICS uses 32-bit signed integers, 
-        # So cast integer values during reading
-        if type(value) is int:
-            value = ctypes.c_int32(value).value
-
-        return value
-
 class EpicsCaServer(object):
     """
     Class to contain an epics ca server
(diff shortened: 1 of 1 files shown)


slacrherbst:
Wouldn't it be better to cast the value from a UInt to an Int in the _variableUpdated method. That way it gets changed before it goes into the database. This would also eliminate the read override, lessening the entries into python when the external clients access any variables.

bengineerd:
What are we going to do when we have a >32-bit unsigned int? The cast currently being done breaks if base=pr.UInt but bitSize>32.

slacrherbst:
I think we need a mechanism to detect when the native type is larger than 32-bits and treat it as a string in epics. There can be a flag in the pvMap which indicates this. This flag would then cause the epics interface to use getDisp() in _variableUpdated in these cases.

jesusvasquez333:
I open another ticket for variables >64bits ESROGUE-107. I think we can break the value into 32-bit words and presented as a waveform PV. But, client will need to pack/unpack the value

slacrherbst:
My preference would be to do the string conversions. It would work in both directions and be easier for the client.

jesusvasquez333:
@slacrherbst: I agree, casting before going to the database is better. But they have to be cast also during _addPV. Otherwise, only the values with a poll interval would work. 

@bengineerd: Thanks for the comment. I added the checking of RemoteVariable too.

slacrherbst:
Are you still making updates or are we stable on this pull request. ?

jesusvasquez333:
It is stable, you can go ahead and merge this PR...

PR #100, opened 2017-08-08
Handle pyro client disconnect

slacrherbst:


python/pyrogue/_Root.py
```diff
@@ -284,12 +285,16 @@ def setTimeout(self,timeout):
...
     def _updateVarListeners(self, yml, l):
-        """Send update to listeners"""
+        """Send update to listeners. Lock must be held."""
         for tar in self._varListeners:
-            if isinstance(tar,Pyro4.Proxy):
-                tar.rootListener(yml,l)
-            else:
-                tar(yml,l)
+            try:
+                if getattr(tar,'rootListener',None) is not None:
```
jesusvasquez333: Wouldn't be better to do?:
`if hasattr(tar,'rootListener'):`

python/pyrogue/_Variable.py
```diff
@@ -184,16 +187,20 @@ def updated(self, var=None, value=None, disp=None):
...
-            else:
-                func(self,value,disp)
+        with self._listenLock:
+            for func in self.__listeners:
+                if value is None:
+                    value = self.value()
+                    disp  = self.valueDisp()
+
+                try:
+                    if getattr(func,'varListener',None) is not None:
```
jesusvasquez333: Wouldn't be better to do?:
`if hasattr(func,'varListener'):`

slacrherbst: Yes you are right.

PR #101, opened 2017-08-09
Add more c++ API calls, add ability to get variable list

slacrherbst:


bengineerd:
This is probably the better way to do it and could actually matter when building large lists:
```
for key,value in self._nodes.items():
    if isinstance(value,pr.BaseVariable):
         lst.append(value)
    else:
        lst.extend(value.variableList)
```

PR #102, opened 2017-08-10
Mysql Support

slacrherbst:


slacrherbst:
We are going to need this for LZ and I don't have time to look into a proper sqlachemy solution. I think we should merge.

PR #104, opened 2017-08-17
Allow pythonic construction of pyrogue.Root()

abunimeh:
By adding __enter__ and __exit__ one can do things like:

with pyrogue.Root(name='ePixBoard', description='ePix10ka Board') as ePixBoard:
	...
	ePixBoard.add(fpga)
	...
	# ePixBoard.stop()  # no need to call stop anymore


slacrherbst:
I lie this  but we don't want to encourage this style of root creation. Instead root should always be subclassed. But the example would be:

with EvalBoard():
    run gui mail loop, etc

and then the stop method would be called.

PR #115, opened 2017-12-05
optional configuration dumps

slacrherbst:


bengineerd:
Use `self.root` instead of `self._root`.

slacrherbst:
would it not be better for performance to use a direct access versus a method? I guess we consider this private and bad practice so better to use self.root?

bengineerd:
Right. _root is private. We should use the public accessors when we have them.

jesusvasquez333:
I will give this try...

jesusvasquez333:
The changes work correctly...

PR #121, opened 2018-01-11
Add GIL acquire in rstTransaction.

slacrherbst:


bengineerd:
Seems pretty straightforward. I'll try to run it as soon as I can.

slacrherbst:
So far it is rock solid with me. I am going to merge it.

slacrherbst:
The only thing I don't like about it is it causes a sequence of release, acquire, release acquire when rstTransaction gets accessed from the python level. This does not occur very often though.

PR #126, opened 2018-01-18
Updated edit fields

slacrherbst:


slacrherbst:
Yes your are correct. I will correct the init line to take advantage of that. The way it was previously written was not working.

bengineerd:
Or more specifically, when extending Device, it's probably better for all the args that will be passed straight to the Device __init__ to be rolled into **kwargs.

bengineerd:
DataWriter could have `hidden` rolled into `**kwargs` as well.

slacrherbst:
I want a different default value for hidden for DataWriter and RunControl

bengineerd:
Ahh, ok.

PR #129, opened 2018-01-18
Added default value for commands.

slacrherbst:


slacrherbst:
When LocalCommands are created the user passes a value field to define the arg type. This value is now used to initialize the command arg.

PR #134, opened 2018-01-18
Improved Version Tracking, auto maintenance increment

slacrherbst:
The maintenance field of the version will now be incremented by the relative commit offset. The full offset information is now printed at rogue startup:

In [1]: import rogue;
Rogue/pyrogue version v2.4.0-1-gdaa8319. https://github.com/slaclab/rogue
In [2]: rogue.Version.major()
Out[2]: 2
In [3]: rogue.Version.minor()
Out[3]: 4
In [4]: rogue.Version.maint()
Out[4]: 1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,15 +20,12 @@
 # ----------------------------------------------------------------------------
 
 
-VER_MAJOR := $(shell git describe --tags | awk -F'[Vv.-]' '{print $$2}')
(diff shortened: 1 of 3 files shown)


slacrherbst:
The maintenance field of the version will not be incremented by the relative commit offset. The full offset information is now printed at rogue startup.

PR #138, opened 2018-01-27
Updates to how var updates are handled

slacrherbst:
While working on ESROGUE-151 (pyro gui performance) I found some items which are hurting our performance when propagating variable updates. I am proposing a few changes which may impact existing code. 

1. The varUpdate callback from the variable changes format from:
   func(var,value,disp)
to
  func(path,value,disp)

The path string is passed instead of the variable object. Passing an object has overhead with pyro (also creates infinite loops) and most receivers just need the path. The local name or object can be obtained through additional local calls. i.e. self.root.getNode(path)

2. The root level variable update callback changes from:
   func(xml, list)
to
   func(path,value,disp)

Basically it becomes exactly the same as the variable level callback. The only different now is that root level call back receivers get all variable changes, where going right to the variable allows a receiver to register for targeted callbacks. As far as I can tell my Rogue internals are the only code which was using the root level callbacks.

I also separated the xml stream generation from the callbacks. Now the callbacks are immediate, while the xml generate can be blocked during reads or writes. This reduced the number of objects that I was storing and generating.

I also removed some unnecessary locks after reading up on python collection thread safety.

Let me know if you know of code that is using the passed var object in the variable update listener or code which was using the root level variable callback.

### DIFF
--- a/python/pyrogue/_Root.py
+++ b/python/pyrogue/_Root.py
@@ -203,7 +203,7 @@ class Root(rogue.interfaces.stream.Master,pr.Device):
     def addVarListener(self,func):
         """
         Add a variable update listener function.
-        The variable, value and display string will be passed as an arg: func(var,value,disp)
+        The variable, value and display string will be passed as an arg: func(path,value,disp)
         """
         self._varListeners.append(func)
 
@@ -386,23 +386,23 @@ class Root(rogue.interfaces.stream.Master,pr.Device):
             self.SystemLog.set(value='',write=False)
         self.SystemLog.updated()
 
-    def _varUpdated(self,var,value,disp):
+    def _varUpdated(self,path,value,disp):
         for func in self._varListeners:
             if hasattr(func,'varListener'):
-                func.varListener(var,value,disp)
+                func.varListener(path,value,disp)
             else:
-                func(var,value,disp)
+                func(path,value,disp)
 
         with self._updatedLock:
 
             # Log is active add to log
             if self._updatedDict is not None:
-                addPathToDict(self._updatedDict,var.path,disp)
+                addPathToDict(self._updatedDict,path,disp)
 
             # Otherwise act directly
             else:
                 d   = {}
-                addPathToDict(d,var.path,disp)
+                addPathToDict(d,path,disp)
                 yml = dictToYaml(d,default_flow_style=False)
                 self._sendYamlFrame(yml)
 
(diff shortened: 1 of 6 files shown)


jesusvasquez333:
@slacrherbst: I tried this using the [ESROGUE-151 branch](https://github.com/slaclab/rogue/tree/ESROGUE-151). I don't get exceptions anymore, however the tree view in the remote gui is empty. I just see the "enable" variable at the root level, and only the root level commands.

slacrherbst:
@jesusvasquez333 Interesting. The way it works now is the GUI will populate as entries are received over Pyro. If it stays in that state it means it is not connecting and iterating through the modules. Are there any error messages appearing in either console?

jesusvasquez333:
@slacrherbst When I start the remote client a was able to read variables (I tried for example read the UpTimeCnt in the AxiVersion module) from the python shell. But When I started the GUI is was empty. So the connection was there, but for some reason the GUI was not populated.

I didn't see any error on the console.

I did something like this in the client side:

```
[jvasquez@tid-pc93130 dev]$ python3                                                                                              
Python 3.6.1 (default, May 19 2017, 14:27:12)                                                                                    
[GCC 4.4.7 20120313 (Red Hat 4.4.7-18)] on linux                                                                                 
Type "help", "copyright", "credits" or "license" for more information.                                                           
>>> import sys                                                                                                                   
>>> import pyrogue                                                                                                               
Rogue/pyrogue version 2.4.0. https://github.com/slaclab/rogue                                                                    
>>> import PyQt4.QtGui
>>> import pyrogue.gui
>>> client = pyrogue.PyroClient(group="test_jesus", host="tid-pc93130")
>>> root = client.getRoot('AMCc')                                      
>>> root.FpgaTopLevel.AmcCarrierCore.AxiVersion.UpTimeCnt.get()        
4725                                                                   
>>> app_top = PyQt4.QtGui.QApplication(sys.argv)                       
>>> gui_top = pyrogue.gui.GuiTop(group='GuiTop')                       
>>> gui_top.addTree(root)
```

slacrherbst:
@jesusvasquez333 You have to start the QT event handler which is blocking
appTop.exec_()

jesusvasquez333:
@slacrherbst: Sorry, in my previous comment the code snip was incomplete (I edited it and now is correct). I did run the `app_top.exec_()` command at the end.

PR #143, opened 2018-02-06
Packetizer Version 2 support

slacrherbst:


ruck314:
@slacrherbst: Has the software been updated with the latest updates in the firmware Version 2 packetizer?

slacrherbst:
Yes its been sitting in a pull request for a couple of weeks.

ruck314:
The firmware formatting has recently changed.  Does the software reflect these firmware changes?  If yes, then I can start to test this software on the dev board platform.

slacrherbst:
If you mean the version number addition, yes I updated to match that.

ruck314:
Ok, I will try it on test it on the dev board platform now

ruck314:
@slacrherbst: My existing code is this:
```
    ethLink = pr.protocols.UdpRssiPack(
        host = args.ip,
        port = 8192,
        size = 1400,
        )
```
What do I need to change to select whether I am using packetizer V1 or packeter V2?

slacrherbst:
Added a new arg packVer to the wrapper.

ruck314:
@slacrherbst:

ruck314:
@slacrherbst: Is "75df5ff" a bug fix for the issue that I am seeing?

slacrherbst:
No. There must be a protocol mismatch someplace. Let me add debugging.

ruck314:
@slacrherbst: Since the RSSI already has a checksum, we have the option to turn on/off the CRC in the packeterizer.  For high bandwidth performance, should we disable the packeter CRC checking/generation? Or should we assume always CRC in the packerize for RSSI+UDP communication?  I think we should always do one way or the other (and not both).  Having support support two different modes of CRC will get confusing fast.

ruck314:
I would vote to "always" do the CRC in the packetizer (another level of error checking).

slacrherbst:
Agreed.

slacrherbst:
To Enable packetizer debugging:
rogue.Logging.setFilter("packetizer.Controller",rogue.Logging.Debug)

slacrherbst:
I just pushed a version with more debugging. See previous comment to enable debug.

ruck314:
@slacrherbst: After these modification in the SURF pull request (https://github.com/slaclab/surf/pull/178), I am now able to do register transaction. 
I am now testing interleaved traffic (register + PRBS TX).  I have observed that if the PRBS TX packet is larger than the chucking size that PRBS TX packets are getting dropped.

slacrherbst:
This is the software side sending packets that are too large?

What did you set for the size value on the software side?

slacrherbst:
Can you point me to the test script?

ruck314:
Here's the image:

slacrherbst:
Ok. so you are transmitting from firmware to software? This has to be a firmware error then? I thought you were talking about a firmware transmit.

ruck314:
Yes, firmware-to-software transfer.
I have not tested the PRBS software-to-firmware path yet

slacrherbst:
Nothing gets received in the software for the dest? No Debug? It could still be a software error

slacrherbst:
Please remove setup_env.csh from the commit.

ruck314:
Done.

ruck314:
@slacrherbst: https://github.com/slaclab/rogue/blob/ESROGUE-153/src/rogue/protocols/packetizer/ControllerV2.cpp#L142

slacrherbst:
Fixed it. Bad merge of two conditions.

ruck314:
@slacrherbst: can you check the thread locking of module?  If I have polling turned off, I don't get any drop packets or PRBS errors.  Then when I hit the ReadAll button a bunch of times, I get CRC errors in the packetizer and PRBS errors.

slacrherbst:
Locking looks ok. CRC errors in the packetizer are interesting. That means the packet got corrupted or shortened someplace between the firmware packetizer and the software module.

ruck314:
@slacrherbst: I think that we should wait for Ben to complete PGPv3 simulation regression testing to confirm that FW-to-FW still looks good.

slacrherbst:
Once you get this moving data we should merge it into pre-release.

ruck314:
@slacrherbst: I need to test the path where the software is the PRBS TX and the firmware is PRBS RX.  Can you point me to an example of this?

slacrherbst:
prbTx = pyrogue.utilities.prbs.PrbsTx(name="prbTx")
pyrogue.streamConnect(prbTx, txVc)

You will be able to start/stop it in software.

ruck314:
Completed by software testing of this software branch with respect to the SURF (https://github.com/slaclab/surf/pull/178/commits/5e3942cee2ed28f96e222bfca68320bf09e41c80).  I have tested PRBS in both directions for both single chucked up frames and multiple chucked up frames.  I think we are ready for the merge to pre-release branch.
@slacrherbst: Anything else that you want to test before the merge?  Do you want to resolve the merge conflict?

slacrherbst:
We will still want full CRC mode over Ethernet when going firmware -> firmware, but we should not with software if we want performance. So configurable.

ruck314:
Unless there is going to be a change to the protocol definition, I think that we are ready to merge pull request.

src/rogue/protocols/packetizer/Controller.cpp
```diff
@@ -55,7 +55,8 @@ rpp::Controller::Controller ( uint32_t segmentSize, rpp::TransportPtr tran, rpp:
 
    headSize_ = headSize;
    tailSize_ = tailSize;
-   enCrc_    = false; 
+   enIbCrc_ = false; // Need to bring this up to the top level
```
ruck314: @slacrherbst: Can I have you add the hooks to map this to the higher levels?

slacrherbst:
Ok. Will do.

src/rogue/protocols/packetizer/Controller.cpp
```diff
@@ -55,8 +55,6 @@ rpp::Controller::Controller ( uint32_t segmentSize, rpp::TransportPtr tran, rpp:
 
    headSize_ = headSize;
    tailSize_ = tailSize;
-   enIbCrc_ = false; // Need to bring this up to the top level
```
ruck314: @slacrherbst: Where does the (enIbCrc_ = false) and (enObCrc_ = true) as default in the higher level software?

PR #150, opened 2018-02-13
Large blocks

slacrherbst:


slacrherbst:
@bengineerd Should we pull this and have this as a possible feature?

PR #152, opened 2018-02-22
New EPICs interface for review

slacrherbst:
This is still being tested. I am adding this pull request so the implementation can be reviewed.

The API is the same, with the addition of a slave interface to map an epics variable to a rogue stream.

List variable types will generate an error but will not longer be needed because of the stream interface.


ruck314:
@slacrherbst: Do these changes make it required to have EPICS installed if you want to build rogue?

slacrherbst:
The makefile changes make it possible to build without EPICS installed. 
I am considering adding the same feature for the aes-drivers if you are 
doing a pure network install of rogue.

slacrherbst:
Just made other change which will break backward compatibility. 
I think we should move epics into protocols/
I moved protocols.py to protocols/_Network.py and auto include its contents in protocols/__init__.py
epics will now be in protocols/epics/....

slacrherbst:
The new epics interface is ready!

PR #153, opened 2018-02-23
Move smem, simulation and mysql to interfaces directory

slacrherbst:
I also moved protocols.py to protocols/_Network.py and auto included from protocols/__init__.py

Usage of UdpRssiPack won't change.


slacrherbst:
Please review and comment.

PR #159, opened 2018-03-02
Fix SSI Default and Error Handling In DataCard and AxiStream Interface

slacrherbst:


ruck314:
After these changes, is the default behavior the following?
1) Add SOF flag on outboard by default (enSsi_=true)
2) Check for SOF flag and EOFE flag on inbound by default (enSsi_=true)

slacrherbst:
>  1. Add SOF flag on outboard by default (enSsi_=true)

Yes

>  2. Check for SOF flag and EOFE flag on inbound by default (enSsi_=true)

Yes on EOFE, no on SOF. It does not do SOF encforcement on the way in.

PR #160, opened 2018-03-02
Update Buffer to fix bugs and corner cases

slacrherbst:
This pull request contains the changes in PR #143 and must not be merged until after PR #143 has been merged and tested. We will then regression test the changes in the PR.

In fact it may be confusing to look at the changes before PR #143 has been merged.


include/rogue/protocols/rssi/Header.h
```diff
@@ -49,7 +49,7 @@ namespace rogue {
             public:
 
                //! Header Size
-               static const uint32_t HeaderSize = 8;
+               static const  int32_t HeaderSize = 8;
```
ruck314: Why changed from uint32 to int32?

slacrherbst: Because it is passed to an arg which takes a relative adjustment: 

 frame->getBuffer(0)->adjustHeader(-rpr::Header::HeaderSize);

slacrherbst: Wanted to avoid singed/unsigned confusion on that line.

PR #161, opened 2018-03-02
 Added sprintf style creator for general error

slacrherbst:


include/rogue/GeneralError.h
```diff
@@ -33,10 +33,11 @@ namespace rogue {
     * in the system.
     */
    class GeneralError : public std::exception {
-         char text_[200];
+         char text_[600];
```
ruck314: Should "600" be defined as a constant in a header file?  It appears to get used again GeneralError.cpp

slacrherbst: I knew one of you guys would go after that :). I will do it properly.

PR #163, opened 2018-03-02
Added new axi hardware

slacrherbst:


ruck314:
Is there an assumption that all transaction in this class is always 4 bytes?

slacrherbst:
Yes. The driver and architecture assume this. This is the same limitation in both the existing DataMap and MapMemory classes.

PR #172, opened 2018-03-13
Add proper iterator for stream interface and add transaction class for memory interface

slacrherbst:
### Stream Iterator

This first change in this pull request modifies the Frame/Buffer interface to use proper iterators. The existing `frame->getCount()` and `frame->getBuffer()` are now replaced with `frame->beginBuffer()` and `frame->endBuffer()` iterators. The previous iterator model for frames (`startRead`, `startWrite`, `nextRead`, `nextWrite`) have been removed in favor of a proper frame iterator in the form of `frame->begin()` and `frame->end()`.

The buffer interface to data is now also consistent with the iterator model allowing the user to access buffer bytes in the following ways:
```c++
// for payload data space
for (Buffer::iterator it = buffer->begin(); it != buffer->endPayload(); ++it)  
```
```c++
// for all frame data space
for (Buffer::iterator it = buffer->begin(); it != buffer->end(); ++it) 
```

Users can choose to iterate directly through the entire frame byte array in the following ways:
```c++
// for frame payload data space
for (Frame::iterator it = frame->begin(); it != frame->endPayload(); ++it)  
```
```c++
// for all frame data space
for (Frame::iterator it = frame->begin(); it != frame->end(); ++it) 
```
Alternatively a user can iterate over buffers and treat the buffer data as large blocks:
```c++
for(Frame::BufferIterator it = frame->beginBuffer(); it != frame->endBuffer(); ++it)
   write(socket, (*it)->begin(), (*it)->endPayload()-(*it)->begin());  
```
Both the frame and buffer have cleaner methods to get and adjust frame lengths. The python interface to frame data does not change.

### Memory Transaction Class
The second change is a major modification to the memory interface. A new class type `Transaction` is created to contain the transaction information as well as provide the interface to moving data and completing the transaction. This new class is passed between the master and slave to initiate a memory transaction. The slave then operates on this transaction object to get and put data and complete the transaction. Right now this transaction wraps a `char *` iterator, but in the future more complex iterators will be added as needed by more complicated memory transaction masters. 

Both python and c++ memory slave code requires updates to work with this new model. Python memory slaves have a cleaner interface now when interfacing with the `Transaction` object.

### DIFF
--- a/src/rogue/hardware/rce/MapMemory.cpp
+++ b/src/rogue/hardware/rce/MapMemory.cpp
@@ -22,7 +22,9 @@
 #include <rogue/hardware/rce/MapMemory.h>
 #include <rogue/interfaces/memory/Constants.h>
 #include <rogue/GeneralError.h>
+#include <rogue/GilRelease.h>
 #include <boost/make_shared.hpp>
+#include <boost/thread.hpp>
 #include <stdio.h>
 #include <unistd.h>
 #include <sys/types.h>
@@ -114,6 +116,9 @@ void rhr::MapMemory::doTransaction(rim::TransactionPtr tran) {
    if ( (tran->size() % dataSize) != 0 ) tran->done(rim::SizeError);
 
    count = 0;
+
+   rogue::GilRelease noGil;
+   boost::unique_lock<boost::mutex> lock(tran->lock);
    it = tran->begin();
 
    if ((ptr = findSpace(tran->address(),tran->size())) == NULL) {
@@ -134,6 +139,7 @@ void rhr::MapMemory::doTransaction(rim::TransactionPtr tran) {
    }
 
    log_->debug("Transaction id=0x%08x, addr 0x%08x. Size=%i, type=%i, data=0x%08x",tran->id(),tran->address(),tran->size(),tran->type(),data);
+   lock.unlock();
    tran->done(0);
 }
 
--- a/src/rogue/hardware/rce/MapMemory.cpp
+++ b/src/rogue/hardware/rce/MapMemory.cpp
@@ -137,9 +137,9 @@ void rhr::MapMemory::doTransaction(rim::TransactionPtr tran) {
       count += dataSize;
       it    += dataSize;
    }
+   lock.unlock(); // Done with iterator
 
    log_->debug("Transaction id=0x%08x, addr 0x%08x. Size=%i, type=%i, data=0x%08x",tran->id(),tran->address(),tran->size(),tran->type(),data);
-   lock.unlock();
    tran->done(0);
 }
 
--- a/src/rogue/hardware/rce/MapMemory.cpp
+++ b/src/rogue/hardware/rce/MapMemory.cpp
@@ -101,27 +101,40 @@ uint8_t * rhr::MapMemory::findSpace (uint32_t base, uint32_t size) {
 }
 
 //! Post a transaction
-void rhr::MapMemory::doTransaction(uint32_t id, boost::shared_ptr<rogue::interfaces::memory::Master> master, 
-                                   uint64_t address, uint32_t size, uint32_t type) {
-   uint8_t * ptr;
+void rhr::MapMemory::doTransaction(rim::TransactionPtr tran) {
+   rim::Transaction::iterator it;
+
    uint32_t count;
+   uint32_t data;
+   uint32_t dataSize;
+   uint8_t * ptr;
 
-   if ((ptr = findSpace(address,size)) == NULL) {
-      master->doneTransaction(id,rim::AddressError);
+   dataSize = sizeof(uint32_t);
+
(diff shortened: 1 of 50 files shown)


src/rogue/hardware/rce/MapMemory.cpp
```diff
@@ -101,27 +103,44 @@ uint8_t * rhr::MapMemory::findSpace (uint32_t base, uint32_t size) {
...
+         std::copy(ptr,ptr+count+dataSize,it);
+      }
+      ptr   += dataSize;
+      count += dataSize;
+      it    += dataSize;
    }
+
+   log_->debug("Transaction id=0x%08x, addr 0x%08x. Size=%i, type=%i, data=0x%08x",tran->id(),tran->address(),tran->size(),tran->type(),data);
+   lock.unlock();
+   tran->done(0);
```
ruck314: To confirm: Does putting the tran->done(0) after the lock (instead of before make sense)?

slacrherbst: Yes. It is neccessary to avoid deadlocks.
The lock is there to protect the memory pointers while the iterator is being used.
tran->done is thread safe

src/rogue/interfaces/stream/Slave.cpp
```diff
@@ -70,11 +73,13 @@ void ris::Slave::acceptFrame ( ris::FramePtr frame ) {
...
-         frame->read(&val,x,1);
+      count = 0;
+      for (it = frame->begin(); (count < debug_) && (it != frame->end()); ++it) {
+         count++;
+         val = *it;
 
          snprintf(buffer + strlen(buffer),1000-strlen(buffer)," 0x%.2x",val);
-         if (( (x+1) % 8 ) == 0) {
+         if (( (count+1) % 8 ) == 0) {
             log_->log(100,buffer);
```
ruck314: ```log_->log(100,......);```
Isn't there a header constant that you should use instead of hard coding to 100?

slacrherbst: That should probably just be critical.

src/rogue/protocols/packetizer/ControllerV1.cpp
```diff
@@ -183,37 +183,38 @@ void rpp::ControllerV1::applicationRx ( ris::FramePtr frame, uint8_t tDest ) {
...
+      size = (*it)->getPayload();
+      data = (*it)->begin();
 
       data[0] = ((appIndex_ % 16) << 4);
       data[1] = (appIndex_ / 16) & 0xFF;
 
-      data[2] = x % 256;
-      data[3] = (x % 0xFFFF) / 256;
-      data[4] = x / 0xFFFF;
+      data[2] = segment % 256;
```
ruck314: Using a "%" operator is not very CPU efficient and hurts us in streaming performance. 
I would recommend implementing with bit operators instead (see what I did in the ControllerV2.cpp code).

src/rogue/protocols/packetizer/ControllerV2.cpp
```diff
@@ -238,25 +240,25 @@ void rpp::ControllerV2::applicationRx ( ris::FramePtr frame, uint8_t tDest ) {
...
 
-   for (x=0; x < frame->getCount(); x++ ) {
+   segment = 0;
+   for (it=frame->beginBuffer(); it != frame->endBuffer(); ++it) {
       ris::FramePtr tFrame = ris::Frame::create();
-      buff = frame->getBuffer(x);
 
       // Compute last, and alignt payload to 64-bits
-      last = buff->getPayload() % 8;
+      last = (*it)->getPayload() % 8;
```
ruck314: ```last = (*it)->getPayload() & 0x7;``` is faster.

src/rogue/protocols/packetizer/ControllerV1.cpp
slacrherbst: Ok. That can be addressed in the current pre-release branch.

src/rogue/utilities/fileio/StreamReader.cpp
```diff
@@ -165,20 +169,34 @@ void ruf::StreamReader::runThread() {
                break;
             }
 
+            // Skip next step if frame is empty
+            size -= 4;
```
ruck314: is it possible for size to not be a multiple of 4?  There's no checks for it.

src/rogue/hardware/rce/MapMemory.cpp
ruck314: Ok, just making sure

src/rogue/utilities/fileio/StreamReader.cpp
slacrherbst: Yeah. That should be: 
if ( size <= 4 ) continue;
size -= 4;

PR #173, opened 2018-03-13
Allow Hub.doTransaction() to be overridden in Python

bengineerd:
A pyrogue Device can now override `_doTransaction()` to perform a limited amount of address manipulation.


slacrherbst:
This will be a messy merge with #172 Can you point me to where this is used right now?

python/pyrogue/_Device.py
```diff
@@ -400,7 +400,7 @@ def _rootAttached(self, parent, root):
         pr.Node._rootAttached(self, parent, root)
 
         self._maxTxnSize = self._reqMaxAccess()
-        self._minTxnSize = self._reqMinAccess()
```
slacrherbst: We may as well do both Min and Max?

PR #174, opened 2018-03-13
Variable default value enhancements

bengineerd:
Fixed RemoteVariable initial/default values

Can now also give a Device a dict of default values for its Variables

### DIFF
--- a/python/pyrogue/_Block.py
+++ b/python/pyrogue/_Block.py
@@ -20,7 +20,6 @@ import math
 import textwrap
 import pyrogue as pr
 import inspect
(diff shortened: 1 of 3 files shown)


python/pyrogue/_Block.py
```diff
@@ -164,7 +164,7 @@ def stale(self):
     def set(self, value):
         with self._lock:
             changed = self._value != value
-            self._value = copy.copy(value)
```
slacrherbst: We should leave this to avoid locking issues with lists.

PR #178, opened 2018-03-15
Generate address space overlap warnings

slacrherbst:
This pull request adds two features. 

The first looks for overlapping variables within a block. This is done by doing a bit by bit check as variables are added to the block. A critical log message is generated when an overlap is detected.

The second feature looks for overlapping device address spaces.  In order for this feature to work properly I added a unique ID to each memory interface slave. The device will query the memBase chain to find the ID of the memory interface it is connected to. This allows the pyrogue root to avoid mislabeling overlaps between devices connected to different memory interfaces. If any devices overlap and are connected to the same memory interface, a critical log message is generated.

### DIFF
--- a/include/rogue/interfaces/memory/Hub.h
+++ b/include/rogue/interfaces/memory/Hub.h
@@ -58,6 +58,9 @@ namespace rogue {
                //! Get offset
                uint64_t getOffset();
 
+               //! Return ID to requesting master
+               uint32_t doSlaveId();
+
                //! Return min access size to requesting master
                uint32_t doMinAccess();
 
--- a/include/rogue/interfaces/memory/Master.h
+++ b/include/rogue/interfaces/memory/Master.h
@@ -85,6 +85,9 @@ namespace rogue {
                //! Get slave
                boost::shared_ptr<rogue::interfaces::memory::Slave> getSlave();
 
+               //! Query the slave ID
+               uint32_t reqSlaveId();
+
                //! Query the minimum access size in bytes for interface
                uint32_t reqMinAccess();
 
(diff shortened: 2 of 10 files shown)


slacrherbst:
I assume we only want warnings and not exceptions?

bengineerd:
I think exceptions might be better. People are really good at ignoring warnings.
If we ever find a case where we want there to be overlaps we can change it to warnings.

ruck314:
I agree.  If they need "overlapping", they can achieve this with LinkVariable.  Exception would prevent this bad practice

jesusvasquez333:
When working with the pre-release I found several overlapping registers. I would say we should generate warning in the next release, and exceptions in the following release. This will help to find and fix problems and a smother transition.

slacrherbst:
Were the overlaps device to device or within a device? Seems that overlaps within a device would be pretty serious, but you may have two devices which play in the same space? The more I think about it the more it scares me that this would even exist.

bengineerd:
Yes, I would very much like to know where you saw these overlaps. If it was in SURF pyrogue modules we should fix them right away.

jesusvasquez333:
Some examples are:

https://github.com/slaclab/lcls-timing-core/blob/4de46d35c3536879b0fb4deff6a70b241a0a67e0/python/LclsTimingCore/TPGMiniCore.py#L83-L101

https://github.com/slaclab/surf/blob/73289c19a6106b4031b137f9af17c079eab07e14/python/surf/devices/ti/_Lmk04828.py#L549-L558 with https://github.com/slaclab/surf/blob/73289c19a6106b4031b137f9af17c079eab07e14/python/surf/devices/ti/_Lmk04828.py#L1070-L1079

many registers in in Daq38J84 overlap with https://github.com/slaclab/surf/blob/73289c19a6106b4031b137f9af17c079eab07e14/python/surf/devices/ti/_Dac38J84.py#L46-L58

And so on...

I'm gathering all the overlaps and I will put then on a Jira ticket soon

PR #182, opened 2018-03-18
Set flags in outbound frame for packetizer

slacrherbst:
When doing another update I noticed that the flags were being modified in the inbound frame instead of the outbound frame.


slacrherbst:
@ruck314 Can you double check this?

PR #183, opened 2018-03-19
Add __version__ attribute with the module version number

jesusvasquez333:
Add the `__version__` attribute with the version number. When importing `pyrogue`the module one can call `pyrogue.__version__` and importing other modules conditionally. An example is the EPICS module which is different in the current release and the after the pre-release branch will be merge.

### DIFF
--- a/include/rogue/Version.h
+++ b/include/rogue/Version.h
@@ -55,6 +55,8 @@ namespace rogue {
          static uint32_t getMinor ();
          static uint32_t getMaint ();
          static uint32_t getDevel ();
+
+         static std::string pythonVersion();
    };
 }
 
--- a/python/pyrogue/__init__.py
+++ b/python/pyrogue/__init__.py
@@ -129,4 +129,5 @@ def genBaseList(cls):
     return ret
 
 # Add __version__ attribute with the module version number
-__version__ = rogue.Version.current()
+__version__ = rogue.Version.pythonVersion()
+
--- a/python/pyrogue/__init__.py
+++ b/python/pyrogue/__init__.py
@@ -128,3 +128,5 @@ def genBaseList(cls):
 
     return ret
 
+# Add __version__ attribute with the module version number
+__version__ = rogue.Version.current()
--- a/src/rogue/Version.cpp
+++ b/src/rogue/Version.cpp
@@ -96,6 +96,18 @@ uint32_t rogue::Version::getDevel() {
    return _devel;
 }
 
+std::string rogue::Version::pythonVersion() {
+   init();
+   std::stringstream ret;
+
+   ret << std::dec << _major;
+   ret << "." << std::dec << _minor;
+   ret << "." << std::dec << _maint;
+
+   if ( _devel > 0 ) ret << ".dev" << std::dec << _devel;
+   return ret.str();
+}
+
 void rogue::Version::setup_python() {
    bp::class_<rogue::Version, boost::noncopyable>("Version",bp::no_init)
       .def("current", &rogue::Version::current)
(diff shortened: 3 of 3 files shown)


slacrherbst:
@jesusvasquez333 Is this preferred over using the features included in the rogue.Version package. One of the calls is rogue.Version.minVersion()

jesusvasquez333:
@slacrherbst Even though one could achieve the same calling `rogue.Version.current()`, I think the `__version__` is more standard in python module, and it follows [PEP 396](https://www.python.org/dev/peps/pep-0396/).

Note: I just realize that `rogue.Version.minVersion()` throws an Argument error

```
>>> rogue.Version.minVersion()
  File "<stdin>", line 1, in <module>
Boost.Python.ArgumentError: Python argument types in
    Version.minVersion()
did not match C++ signature:
    minVersion(std::string)
>>>
```

slacrherbst:
min version works like this:

rogue.Version.minVersion("v2.5.0")

 and will throw an exception if that is not the case. Is version setup to do similiar range compares.

I agree we should be more python standard. Maybe I can remove some features from rogue.Version and we just the use the standard as you suggest. Are there limitations in the format of the version? For example rogue.Version.current() can possibly return something like v2.5.0-444-somehash. 

Should I create a call which returns a python compatible format? 

The values we have in our version are major.minor.main.devel where devel is auto generated by the git variance from the current tag.

slacrherbst:
An example would be to create a call like:

rogue.Version.getPythonVersion() which would convert:

v2.5.0-444-g44h7s 
to
2.5.0.444

jesusvasquez333:
@slacrherbst Thanks, I didn't look closer to how to use `rogue.Version.minVersion`. 

In [PEP 396](https://www.python.org/dev/peps/pep-0396) they mention that he version number should follow the normalized version format specified in [PEP 386](https://www.python.org/dev/peps/pep-0386/) which is 

`N.N[.N]+[{a|b|c|rc}N[.N]+][.postN][.devN]`

Some examples given are:

```
1.0a1
1.0a2.dev456
1.0a2
1.0a2.1.dev456
1.0a2.1
1.0b1.dev456
1.0b2
1.0b2.post345
1.0c1.dev456
1.0c1
1.0.dev456
1.0
1.0.post456.dev34
1.0.post456
1.0rc2
```

So, I think `v2.5.0-444-g44h7s` will have to be something like `2.5.0` + `.postXXX` for a post release or `.devXXX` for a pre-release

slacrherbst:
Please see my commit just now for a python compatible version output: 
````
In [1]: import pyrogue
Rogue/pyrogue version v2.6.0-311-g6ecf68d. https://github.com/slaclab/rogue

In [2]: pyrogue.__version__
Out[2]: '2.6.0.dev311'
````

**Comment by jesusvasquez333** [comment]
@slacrherbst Is the 311 a number that increments with every commit?

**Comment by slacrherbst** [comment]
It comes directly out of the git describe --tags command. I believe it is the number of commits since the tag.

**Comment by jesusvasquez333** [comment]
@slacrherbst I found this:
```
With something like git.git current tree, I get:

[torvalds@g5 git]$ git describe parent
v1.0.4-14-g2414721
i.e. the current head of my "parent" branch is based on v1.0.4, but since it has a handful commits on top of that, describe has added the number of additional commits ("14") and an abbreviated object name for the commit itself ("2414721") at the end.

The number of additional commits is the number of commits which would be displayed by "git log v1.0.4..parent". The hash suffix is "-g" + 7-char abbreviation for the tip commit of parent (which was 2414721b194453f058079d897d13c4e377f92dc6).
```
So, yes: 311 is the number of commits after the tag.

So, it seems good to me.

PR #185, opened 2018-03-20
bringing pre release changes

jesusvasquez333:


slacrherbst:
Just an FYI. Updates to development branches can be done with simple merges instead of pull requests.

slacrherbst:
It also removes extra clutter from the release notes as well.

PR #188, opened 2018-03-21
Fix bad vector iterator in memory Slave transaction tracking

slacrherbst:
Also includes cleanup of redundant GIL release in epics Slave interface and cleans up transaction deletion in SrpV3 to match SrpV0.


jesusvasquez333:
@slacrherbst I tried this version and I don't get any more set faults. The changes look good.

PR #190, opened 2018-03-26
Add big-endian and fixed-point Models

bengineerd:
**This is still in development but I'd like some feedback on it.**

This PR adds the following new Models:
 * UIntBE - Unsigned big-endian integer
 * IntBe - Signed big-endian integer
 * Float - Existed before but is now explicitly 32-bits
 * Double - An 64-bit float (little-endian)
 * FloatBE - A big-endian Float
 * DoubleBE - a big-endian Double
 * Fixed - A fixed point number

Underneath, there is a big change to the way Models work.
Models are now instantiated with a specific `bitSize`. Model subclasses can also add other parameters, but bitSize is the one that is always required. This change was made mostly to support the fixed-point Models. These need to know both a bitSize and a pointPosition.

There is some special python metaclass magic to make sure that only one instance of each Model is ever instantiated with a given set of parameters. If the code tries to instantiate e.g. UInt(16) twice, the second call with return a reference to the object created by the first call, rather than creating an identical object in memory.

Backward compatibility has been maintained. You can still call:
```python
RemoteVariable(..., base=pr.UInt, bitSize=16, ...)
```
The RemoteVariable constructor will check the `type` of `base`, and if it is a Model class, it will instantiate that class with the `bitSize`. **This is how we should continue to use it.**

It is also now valid to do this:
```python
RemoteVariable(..., base=pr.UInt(16), bitSize=16, ...)
```
**Note that the bitSize param to RemoteVariable() is still necessary. The fact that this works is mostly a side effect of the changes required for base=pr.Fixed to work.**

Fixed-point numbers look like this:
```python
RemoteVariable(..., base=pr.Fixed(16, 15, signed=True), bitSize=16, ...)
```

It's all still a bit clunky (and totally untested) but we should discuss if this is the path we want to go down.


bengineerd:
I will test it now against one of my projects to make sure it works. It might need some refactoring for the latest Rogue.

PR #191, opened 2018-03-27
Add big-endian integer and floating-point Models

bengineerd:
**This PR is an alternative to https://github.com/slaclab/rogue/pull/190. Here we would just add big-endian support but not fixed-point support.**

This PR adds the following new Models:
 * UIntBE - Unsigned big-endian integer
 * IntBe - Signed big-endian integer
 * Float - Existed before but is now explicitly 32-bits
 * Double - An 64-bit float (little-endian)
 * FloatBE - A big-endian Float
 * DoubleBE - a big-endian Double

It also removes Model.mask(), which wasn't doing anything.

This hasn't been tested yet.


slacrherbst:
Will use #190 instead.

PR #194, opened 2018-03-29
Optimize variable update processing

slacrherbst:
This change provides an interface to group variable updates into blocks, with overlap detection in order to minimize the redundant calls to each variable's update listener.

Anytime a variable is written or read a update tracker is created which keeps track of the variables that will need an update notification as a result. After the operation is completed, the update list is handed to a worker thread which will calls the associated update functions for each variable in the list. 

Users may also further limit the update rate by grouping updates outside of a series of calls:

````
with self.root.updateGroup():
     var1.set(value1)
     var2.set(value2)
     var1.set(value3)
````

### DIFF
--- a/python/pyrogue/_Block.py
+++ b/python/pyrogue/_Block.py
@@ -161,7 +161,7 @@ class LocalBlock(BaseBlock):
     def stale(self):
         return False
 
-    def set(self, value):
+    def set(self, var, value):
         with self._lock:
             changed = self._value != value
             self._value = value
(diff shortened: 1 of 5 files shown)


slacrherbst:
One note on this. Right now the with root.updateGroup() call creates a lock and increments and decrements a counter on each respective entry/exit from the with context. Later the calls to _queueUpdate() also gain a lock before adding the variable to the list of pending updates.

An alternative approach would be to post an increment command to the work queue at the entry to the context manager and post a decrement command at the exit. The calls to _queueUpdate() would simply push the variable value onto the work queue. The work thread would then keep track of the entry/exit counts and the variable overlaps. This would remove the necessity to gain a recursive lock multiple times in each transaction.

This would also allow updates from different threads to overlap without blocking.  This change would not impact the API at all.

PR #198, opened 2018-03-29
bug fix for PacketizerV1's SW-to-FW path

ruck314:
### Description
1)  bug fix for PacketizerV1's SW-to-FW path
2) Fixed some typos in the comments
3) Minor performance enhancements (avoiding modulo operator because not high performance)

I have regression tested this with both PacketizerV1 and PacketizerV2.  We still have the PacketizerV1 FW-to-SW corner case (FW bug) that needs to be results (discovered during this development). 

### JIRA
ESROGUE-199
ESCORE-339

### DIFF
--- a/include/rogue/protocols/packetizer/Application.h
+++ b/include/rogue/protocols/packetizer/Application.h
@@ -82,7 +82,7 @@ namespace rogue {
                void acceptFrame ( boost::shared_ptr<rogue::interfaces::stream::Frame> frame );
          };
 
-         // Convienence
+         // Convenience
          typedef boost::shared_ptr<rogue::protocols::packetizer::Application> ApplicationPtr;
 
       }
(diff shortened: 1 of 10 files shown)


slacrherbst:
I will regenerate this as a PR to master later this evening. Some pre-release changes need to be filtered out.

ruck314:
No, it was meant for the pre-release.

ruck314:
I was under the assumption that we stage the patches into the pre-release, do our regression testing in the pre-release, then do our release candidates from pre-release to master.

slacrherbst:
It depends on the fix. This one falls into the grey area. If there is a bug which is known to be a problem and has a very simple fix then it should be applied directly to master and the maint number will increment. Here we have a bug with a simple fix for a problem which is rarely encountered. Let's go ahead and pull into into pre-release.

The idea is to allow quick fixes to known problems to be applied without forcing users to be exposed to more complex fixes and feature adds that will exist in pre-release for a testing period.

PR #199, opened 2018-03-30
Move update context count tracking to worker thread

slacrherbst:
This is a possible enhancement to PR #194.

This version passes the context increment and decrement request as well as the variable to the worker thread through the work queue. The worker thread then tracks the context count as well as the update list. This allows the tracking operation to occur outside of a re-enterent lock context.


slacrherbst:
Github is being annoying and not showing the proper diff to the current state of pre-release. Please see the following link for the relative changes:

https://github.com/slaclab/rogue/compare/pre-release...ESROGUE-200-2

PR #201, opened 2018-04-02
Add stop call to RSSI

slacrherbst:
Add stop call to RSSI. This ensures that the server is put into the reset state before the client disconnects.

The following line will need to be added to the "stop" method of the custom root class or at the end of the python script:

````
rudp.stop()
````

### DIFF
--- a/include/rogue/protocols/rssi/Client.h
+++ b/include/rogue/protocols/rssi/Client.h
@@ -82,6 +82,9 @@ namespace rogue {
                //! Set timeout in microseconds for frame transmits
                void setTimeout(uint32_t timeout);
 
+               //! Stop connection
+               void stop();
+
          };
 
          // Convienence
--- a/include/rogue/protocols/rssi/Server.h
+++ b/include/rogue/protocols/rssi/Server.h
@@ -78,6 +78,9 @@ namespace rogue {
                //! Set timeout in microseconds for frame transmits
                void setTimeout(uint32_t timeout);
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by slacrherbst** [comment]
I have fixed the index error in pre-release and merge the current pre-release into this branch.
As for the timeout, did you add the rupd->stop()?

**Comment by ruck314** [comment]
```

slacrherbst:
There is no good way to detect the rudp class. The reset is also included in the destructor of the rudp class, but it appears that is not guaranteed to be called in a python exit, and never seems to be called in this case. The cleanest way would be to subclass the root device and include the rudp in the stop() method of the sub-class.

ruck314:
what python imports do I need to use for "rupd.stop()"?

slacrherbst:
That was a typo. It was rudp.stop() where rudp is the name of the object you create for the rssi interface. rudp.stop() worked for the eval board project. There would be a need for a check to be sure the rudp  was enabled.

ruck314:
That's kind of the issue that I have in the eval board. I have to pay attention if the RUDP used because dev board support more than RUDP communication.  That's why I'd preferred to have the root.stop() look through the type of communication and do the rudp.stop() if found under the hood.  But it sounds like that's not an option.

slacrherbst:
The easiest thing would be to create root.rudp = None by default. And at the end
````
if root.rudp is not None:
    root.rudp.stop()
````

PR #206, opened 2018-04-11
Fix get for commands

slacrherbst:
Maintenance pull request for ESROGUE-209

The architectural changes to the variable base class ended up breaking the get() method for commands. get() on commands is used to obtain a default arg value.


jesusvasquez333:
@slacrherbst: I just tested this version and the problem is gone

PR #215, opened 2018-05-02
Enable Variable Overlaps

slacrherbst:
This PR provides an interface to enable multiple variables to overlap the same bits in a register space. To enable overlaps all overlapping variables must allow overlaps. The RemoteVariable and RemoteCommand class creators now accept an overlapEn arg.


slacrherbst:
It appears I branched this from the previous PR branch so it has more changes.

PR #220, opened 2018-05-04
Add ability to pass variable list to block methods

slacrherbst:
With this PR in a device you can queue up a list of specific variables for operations using the device block methods. For example to read a list of variables to update a linked variable.

`````
self.readBlocks(variable=linkVar.dependencies)
self.checkBlocks(variable=linkVar.dependencies)
`````

### DIFF
--- a/python/pyrogue/_Device.py
+++ b/python/pyrogue/_Device.py
@@ -405,7 +405,7 @@ class Device(pr.Node,rim.Hub):
 
         blocks = []
         for v in variables:
-            if v not in self:
+            if v.parent is not self:
                 raise DeviceError(
                     f'Variable {v.path} passed to {self.path}._getBlocks() is not a member of {self.path}')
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by bengineerd** [comment]
Should we check that all the Variables passed to `readBlocks` actually belong to the Device?

Also, maybe it's useful to pass the list of Variables to `readBlocks` as strings?

```python
self.readBlocks(variable=[f'Reg{x}' for x in range(10)])
```

slacrherbst:
I am not sure it matters if the variables are in the current device since the blocks are reference directly. As for the string option, most likely this is coming from a dependencies list anyway.

slacrherbst:
Can we just do the following as a check:

`````
for v in variables:
     if v._parent is not self:
`````

bengineerd:
Yes, that is probably better.

PR #222, opened 2018-05-12
Fix SRPV3 Error Handling

slacrherbst:
The SRPV3 module was not properly handling return packets as a result of an AXI Lite bus error.


slacrherbst:
I made the change.

bengineerd:
`TimeoutError` and `BusTimeout` might be a bit confusing.

To clarify: TimeoutError is when there is no SRP response. BusTimeout is when the SRP response indicates a timeout on the downstream bus.

We should keep it as it is and merge this, but these memory error codes need a rethink. Each layer needs to be able to add it's own error codes as a transaction moves though the chain.

slacrherbst:
Agreed. We also need to make sure any error that gets reported has useful information in it. Larry has been finding a few lately that you really need to walk back to figure out what happened.

PR #227, opened 2018-05-15
Additional debug python variable for RSSI

ruck314:
### Description
Additional debug python variable for RSSI


slacrherbst:
Nice use of the lambda

PR #228, opened 2018-05-15
Add augmented arithmetic assignments to local variables

slacrherbst:
Local variables will now support the augmented arithmetic assignments (+=, -=, *=, @=, /=, //=, %=, **=, <<=, >>=, &=, ^=, |=).

### DIFF
--- a/python/pyrogue/_Block.py
+++ b/python/pyrogue/_Block.py
@@ -191,6 +191,71 @@ class LocalBlock(BaseBlock):
     def updated(self):
         self._variable._queueUpdate()
 
+    def __iadd__(self, other):
+        with self._lock:
+            self.set(None, self.get(None) + other)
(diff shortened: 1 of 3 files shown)


bengineerd:
Should there be a public call to acquire a LocalVariable lock?
All of these arith operators are nice, but eventually someone is going to want something more complex.

```python
with self.MyLocalVar.lock:
    value = self.MyLocalVar.get() 
    #some complex operation
    self.MyLocalVar.set( value )
```

I'm also a bit worried that new users won't realize that this locking is required.

slacrherbst:
With the current interfaces external locking is not yet required. We may need it in the future with array and other memory mapped types. The set/get methods are currently called within the lock context. Since there is multiple layers here we will need to create some form of lock token.

PR #230, opened 2018-05-19
RSSI Updates

ruck314:
### Description
* Added oooQueue 
  * Should be rewritten as boost::shared_ptr<rogue::protocols::rssi::Header> oooList_[256];
* Updated the timeout calculations from % operators to div()
  * Recommended by JJ @russell-slac for high performance.  div() is 1 instruction for Intel and 2 for ARM

PR #231, opened 2018-05-22
Added adaptive timeout for memory transactions

slacrherbst:
This PR adds a requirement that memory slave interfaces process transactions in order. The getTransaction() record will now expire old transactions and update the timer for future transactions. delTransaction() is removed.

### DIFF
--- a/include/rogue/interfaces/memory/Slave.h
+++ b/include/rogue/interfaces/memory/Slave.h
@@ -44,10 +44,10 @@ namespace rogue {
                uint32_t id_;
 
                //! Alias for map
-               typedef std::map<uint32_t, boost::weak_ptr<rogue::interfaces::memory::Transaction> > TransactionMap;
+               typedef std::vector<boost::shared_ptr<rogue::interfaces::memory::Transaction> > TransactionList;
 
                //! Transaction map
-               TransactionMap tranMap_;
+               TransactionList tranList_;
 
                //! Slave lock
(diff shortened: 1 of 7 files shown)


ruck314:
Should this PR be for merging into "rssi-busy" branch?  Then we can test it before the https://github.com/slaclab/rogue/pull/230 merge

slacrherbst:
When 1 is received the timer for 2, 3, 4 will be updated. When 3 is received 2 will be declared in error and 4's timer will be updated.

slacrherbst:
If we agree to this PR we will merge it into pre-release and then update #230.

bengineerd:
I think we need to handle the case where a transactions times out, but then arrives next after the timeout before any other transactions.

This seems to happen a lot where an I2C txn response narrowly misses the timeout, but then arrives intact followed by other outstanding responses also intact.

slacrherbst:
In that case the transaction response would be ignored and treated as if it never happened. Its hard to recover since most likely the initiating process would have gone on to other things or re-tried the transaction.

slacrherbst:
The only think "waiting" is the initiating master (python block). It issues all of the transactions and waits on them to timeout. The memory slave will only look for tid=35 if it arrives on the link. Only then does it end the transaction for tid=34. If tid=34 comes late, but before tid=35 the processing calling waitTransaction() will receive the timeout and move on.

PR #232, opened 2018-05-22
Add bit manipulation helpers to memory::Master

slacrherbst:
This PR adds bit field "copy", "set" and "check if set" methods to the memory::Master class to avoid doing bit manipulation for loops at the python level in the Block class.


slacrherbst:
If you comment out 426 - 443 does it get faster?

bengineerd:
Yes. 

I just did a push with the chanMode stuff redone. With the code this way, it starts up instantly.
If I comment that out and uncomment the code below (492-523), it goes back to 1 minute startup time.

slacrherbst:
How about with the chanMode commented out and the 492-523 removed?

PR #233, opened 2018-05-22
Move setTimeout to root.start()

slacrherbst:


bengineerd:
With `timeout=None`, does that mean never time out?

slacrherbst:
Just uses the default, but I guess I should set the default here. Seems like a waste.

bengineerd:
I think setting the default there would be better. It's a bit confusing as-is.

PR #236, opened 2018-05-29
Make writeBlocks() obey force when passed a Variable list

bengineerd:
Fixed a bug where `writeBlocks()` would always force if a list of variables was passed to `variable=`.
It now obeys the `force` parameter.

### DIFF
--- a/python/pyrogue/_Device.py
+++ b/python/pyrogue/_Device.py
@@ -254,13 +254,13 @@ class Device(pr.Node,rim.Hub):
         # Process local blocks.
         if variable is not None:
             for b in self._getBlocks(variable):
-                b.startTransaction(rim.Write, check=checkEach)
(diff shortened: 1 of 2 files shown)


slacrherbst:
You also need to update the set() method in variable to have force=true in writeBlocks

slacrherbst:
Check post as well

bengineerd:
`RemoteVariable.set()` first calls `Block.set()`, which updates the data mask and causes the Block to be stale even if the value didn't change. So it shouldn't matter if force is true or false. I suppose it's a slight optimization though to just call it with `force=True` so it doesn't have to bother checking `Block.stale`.

RemoteVariable.post() doesn't call writeBlocks, so that should be ok.

slacrherbst:
Yes I guess that is correct.

PR #244, opened 2018-06-04
_rawWrite() Update

ruck314:
### Description
adding posted write support to _rawWrite()


slacrherbst:
See update

slacrherbst:
This way it waits for the memory bridge to process before returning. That way if there is a size or address error the error won't be missed. The memory bridge will post then return a completing code immediately.

ruck314:
Looks good to me.  I figured I was missing something. 
Do you want to approve and merge?

PR #246, opened 2018-06-08
Remove deprecated elements

slacrherbst:
Remove hardware/rce and hardware/data.
Remove legacy Variable and Command classes.
Remove legacy command decorators.


slacrherbst:
We should check for any other features/methods that are no longer used, or should not be used.

python/pyrogue/_Command.py
```diff
@@ -219,52 +219,10 @@ def get(self, read=True):
...
-                inspect.getfullargspec(RemoteCommand.__init__).kwonlyargs
-
-        # Pass supported args
-        args = {k:kwargs[k] for k in kwargs if k in cargs}
-
-        ret = RemoteCommand(offset=offset,**args)
-        ret._depWarn = True
-        return(ret)
-
 Command.nothing = BaseCommand.nothing
```
bengineerd: Are these still necessary? Maybe we should just find any code affected and refactor it.

slacrherbst: That section go removed in the PR I thought?

bengineerd:
I think we should move RunControl and DataWriter to utilities.
Or at the very least to their own files?
Maybe not in this PR though, which is more concerned with removing unused code.

slacrherbst:
Maybe but I kind of see them as special devices. Kind of a sub-class which can be customized.

PR #266, opened 2018-08-23
Allow software configurable parameters for RSSI server and client

slacrherbst:
Software may now set the initial negotiation parameters for the RSSI link. With the exception of locMaxSegment (which is defined at class creation) all parameters can be adjusted and are used at the next link initialization.


ruck314:
What testing has been done on this branch?  SW-to-SW only? SW-to-FW only? Or both?

slacrherbst:
I tested sw to fw. It works just as before by default with the new ability to adjust parameters for starting the service.

PR #269, opened 2018-08-29
Source EPICS 3.15 instead of 3.16

jesusvasquez333:


slacrherbst:
Oops. I guess I never updated the .sh version.

PR #270, opened 2018-08-29
Use the EPICS tool EpicsHostArch to get the host architecture

jesusvasquez333:
EPICS base provides a script `EpicsHostArch` which gives the host architecture. The hard-coded `rhel6-x86_64` don't work in other OS, like Ubuntu.


ernestow:
Hi Jesus,


Are you using the version of EPICS that we maintain for the LCLS?

Bruce has already patched EPICS to handle the ARCH


________________________________

PR #271, opened 2018-08-30
Use EPICS_HOST_ARCH variable for architecture instead of epics script

slacrherbst:


jesusvasquez333:
@slacrherbst : what about handling both cases?

slacrherbst:
Looks good. I like it.

PR #274, opened 2018-09-14
Add running status flag to epics server

jesusvasquez333:


slacrherbst:
Ok. I see why. In other code I just use the thread pointer being false to achieve the same thing.

PR #280, opened 2018-10-02
Add option to drop errored frames in file writer

slacrherbst:


jmdewart:
Can we also drop EOFE errors?

frame->getFlags() & 0x100

slacrherbst:
PR #279  will ensure the error field is set when the EOFE flag is set when the packetizer has SSI mode enabled (default setting).

slacrherbst:
And the DMA interface for the PCIE card already has the EOFE -> error feature for SSI.

PR #281, opened 2018-10-02
Add no-copy mode for FIFO

slacrherbst:
This change allows the stream FIFO to operation in a no-copy mode. 

This also attempts to improve the copy efficiency in cases where there are a lot of buffers in the incoming frame.

This changed will break existing code which uses the FIFO due to an additional creation parameter.


jmdewart:
I guess trim and no-copy mode are mutually exclusive.

slacrherbst:
Yes. trim without copy would screw up the data for other receivers.

PR #282, opened 2018-10-03
Added error message when bitsize=0 in remote variables.

slacrherbst:


python/pyrogue/_Variable.py
```diff
@@ -375,6 +375,10 @@ def __init__(self, *,
         if len(offset) != len(bitOffset) != len(bitSize):
             raise VariableError('Lengths of offset: {}, bitOffset: {}, bitSize {} must match'.format(offset, bitOffset, bitSize))        
 
+        # Check for invalid values
+        if 0 in bitSize:
```
ruck314: Does "if 0 in bitSize:" give the same result as "if (bitSize == 0):" in python?

PR #285, opened 2018-10-05
Updated PRBS max width and add max width constant

slacrherbst:


src/rogue/utilities/Prbs.cpp
```diff
@@ -123,7 +123,7 @@ double ru::Prbs::updateTime ( struct timeval *last ) {
 
 //! Set width
 void ru::Prbs::setWidth(uint32_t width) {
-   if ( width != 32 && width != 64 && width != 128 )
+   if ( width < 32 || width > (MaxBytes*8) || (width % 32) != 0)
```
ruck314: The width < 32 is not really necessary due to the modulus check you do in the same line.  Maybe rewrite it as this:
```
if ( width > (MaxBytes*8) || (width % 32) != 0)
```

slacrherbst: Yes I added the 32-bit alignment check later.

PR #287, opened 2018-10-11
Remove aes-stream-drivers as submodule

slacrherbst:
I have decided to remove the aes-stream-drivers as a submoule to simplify the release process and building as an anaconda recipe. Only three headers files are required to be copied from aes-stream-drivers and they change very rarely.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,5 +1,0 @@
-[submodule "drivers"]
-	path = drivers
-	url = https://github.com/slaclab/aes-stream-drivers
-   branch = master
-   ignore = dirty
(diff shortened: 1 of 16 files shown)


slacrherbst:
This includes the changes in the previous pull request due to a branch mistake.

slacrherbst:
https://github.com/slaclab/rogue/blob/51421d40349af34d71db652060ccf5cd0aafd6d4/src/rogue/hardware/axi/AxiStreamDma.cpp#L59-L60

Same as before

bengineerd:
This branch has the YAML fixes in it. We either want to merge that branch first or recreate this branch without the YAML changes.

PR #288, opened 2018-10-15
New build system and anaconda package generation

slacrherbst:
This PR adds support for automatic anaconda package generation and updates the build system to support three possible install approaches:

local: Rogue is compiled and installed in the local checkout directory. Compatible with previous releases.

custom: Rogue is installed in a separate custom directory. i.e. /usr/local/rogue/

system: Rogue is installed in the normal system directories with the python packages installed as a standard python package. This flow is also used for the anaconda builds and may be used for docker builds (untested).

In all cases, except the system install, the setup scripts are auto generated to match the installed directories.

This PR also improves the exported CMake package file to allow third party compiles in all three install situations, including anaconda, while requiring minimal CMake directives in the user's cmake file.


slacrherbst:
I have updated this PR with new Readme files. Please review and approve if you agree.

ruck314:
image
Just try the "bash Anaconda3-5.3.0-Linux-x86_64.sh" instructions on a fresh install of Ubuntu 18.04.  I did noticed that git and git-lfs was the only things that I had to manually install to clone the github project and run rogue (which I don't think is a dependence for rogue).

slacrherbst:
Would it be worth adding instructions for tarball downloads of rogue?

slacrherbst:
Did you do the build version or the direct anaconda download?

slacrherbst:
I mean instead of git download of rogue. When you reference is a separate questions.

PR #289, opened 2018-10-17
Removing mysqlclient from package install instructions

slacrherbst:
Removing mysqlclient from package install instructions


ruck314:
I would add "pip3 install numpy" instructions.  It's include in the conda install but have to do it manually in the build from scratch procedure

slacrherbst:
Done.

PR #301, opened 2018-10-31
Add EPICS to Docker container & use system build for Docker

slacrherbst:
This PR adds an EPICS base build to the docker file and changes the rogue install type to be "system", putting rogue in the standard library directories.

I also removed recordclass from the pip requirements.

The python packages matplotlib, numpy, pyepics are also now installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,15 +7,12 @@ RUN apt-get update && apt-get -y    install wget git \
                                     libreadline6-dev
 
 # Install EPICS
-WORKDIR /
-RUN mkdir epics
-WORKDIR epics
-RUN mkdir base-3.15.5
-WORKDIR base-3.15.5
+RUN mkdir -p /usr/src/epics/base-3.15.5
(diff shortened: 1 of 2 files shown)


jesusvasquez333:
branch is an argument passed from travis depending if we are working on pre-release or master [here](https://github.com/slaclab/rogue/blob/docker_adjust/.travis.yml#L104)

jesusvasquez333:
I build a docker images using this Dockerfile and everything seen correct.

slacrherbst:
> branch is an argument passed from travis depending if we are working on pre-release or master [here](https://github.com/slaclab/rogue/blob/docker_adjust/.travis.yml#L104)

Yes. Thank you.I was testing locally and forgot to put that back.

slacrherbst:
> To be consistent let's install EPICS base in /usr/src

I was not sure what the best directory would be considering it then operates within the directory that it was built in.

PR #306, opened 2018-11-21
Add scipy to conda recipe

bengineerd:
I think scipy has enough useful data analysis tools that we should have it as part of the default rogue env.

Separately, we should write up a procedure for how a project should add it's own conda dependencies on top of the rogue ones.

### DIFF
--- a/conda.yml
+++ b/conda.yml
@@ -17,9 +17,11 @@ dependencies:
   - click
   - pyro4
   - numpy
+  - scipy  
   - pyqt
   - pyepics
   - boost
   - bzip2
   - matplotlib
   - epics-base=3.14.12.8
+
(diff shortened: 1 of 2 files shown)


slacrherbst:
I have two comments on this. First this would only impact the people who are doing rogue builds and not the ones downloading with anaconda. We would need to change the anaconda dependencies for that.

Second I think we should limit "suggested" packages. I am ok with numpy since we will probably do some numpy integrations. Maybe we should consider removing matplotlib.

slacrherbst:
Closing this. We will not add scipy and I am removing matplotlib. Only dependent libraries should be included. numpy will be left as it will be a future dependency.

PR #308, opened 2018-11-28
Allow commands to be run in background threads

slacrherbst:
This pr adds the background attribute to the command class. If this attribute is set to true the command will be executed in the background, with the exec() call returning immediately. This allows for long duration commands to be supported in rogue without blocking the calling process (i.e. epics or GUI).  A second attempt to call the command while the first call is still in progress will result in an immediate return without re-starting the command. 

self.add(Command(name='test', background=True))

### DIFF
--- a/python/pyrogue/_Command.py
+++ b/python/pyrogue/_Command.py
@@ -20,6 +20,7 @@ from collections import OrderedDict as odict
 import pyrogue as pr
 import inspect
 import Pyro4
+import threading
 
 class CommandError(Exception):
     """ Exception for command errors."""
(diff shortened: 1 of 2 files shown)


bengineerd:
Would it be better to have a `BackgroundCommand` class instead? Is there a case where sometimes you'd want to call a command in the foreground and sometimes not?

bengineerd:
From an API standpoint, I worry about having boolean arguments on functions: https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

Our coding style with rogue is to always call with named params though, so I guess that mitigates it a bit.

But also, only allowing LocalCommands to be backgrounded would be OK I think.

PR #322, opened 2019-01-17
Generate error and display INV for invalid enum values

slacrherbst:
This PR will generate the string 'INV' if the integer value received from hardware does no match a known ENUM key.

### DIFF
--- a/python/pyrogue/_Variable.py
+++ b/python/pyrogue/_Variable.py
@@ -243,7 +243,11 @@ class BaseVariable(pr.Node):
         try:
             #print('{}.genDisp(read={}) disp={} value={}'.format(self.path, read, self.disp, value))
             if self.disp == 'enum':
-                ret = self.enum[value]
+                if value in self.enum:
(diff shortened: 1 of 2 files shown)


slacrherbst:
Agreed. Any suggestions as to a good string to pass? I struggled coming up with something:

**INV**?

ruck314:
'INV' usually means "inverter" in firwmare.  I think spelling out "INVALID" makes more sense.

jesusvasquez333:
`**INV**` is better. Or something like `<INV_VAL>`, `<INVALID_VALUE>`, `<INVALID>`, `???`?

bengineerd:
Might be useful for debugging to make it `f'INVALID: {value:x}'`

ruck314:
I like Ben's suggest because it gives feedback of where the issue is

slacrherbst:
I added one more change to better handle this case in a comb box. If the enum index is invalid the combobox will revert to the first selection. Otherwise a bad error will be generated.

PR #328, opened 2019-01-31
Add worker threads to epics interface

slacrherbst:
This PR adds worker threads to the epics interface. Transactions now are handled by separate worker threads instead of the main epics service loop. This allows multiple epics transactions to operate in parallel. The epics class in pyrogue how has a threadCount attribute which defines the number of threads to start.


slacrherbst:
I also had to tweak the epics test to make sure the epics server stopped between each of the loops. I changed a couple of the tests to remove the __enter__ and __exit__ methods since they are already defined in the Root class.

PR #329, opened 2019-01-31
Add forceCheckEach property to Device

bengineerd:
This change adds a property to Device that will force checkEach=True when calling `writeBlocks` , `readBlocks` and `verifyBlocks`.

The idea is that there should be an easy way to tell a Device subclass to always use checkEach=True. Having to override `writeBlocks`, `readBlocks` and `verifyBlocks` is somewhat cumbersome. Now you can just add `self.forceCheckEach = True` in the __init__() method.

As implemented it also forces `checkEach=True` in recursive `*Blocks` calls down the tree. This may or may not be desirable. I have to think about it.

### DIFF
--- a/python/pyrogue/_Device.py
+++ b/python/pyrogue/_Device.py
@@ -125,6 +125,8 @@ class Device(pr.Node,rim.Hub):
         self._blockSize = blockSize
         self._defaults  = defaults if defaults is not None else {}
 
+        self.forceCheckEach = False
+
         # Connect to memory slave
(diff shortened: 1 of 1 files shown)


bengineerd:
Another small tweak also snuck in here:
`_getBlocks()` now effectively ignores LinkVariables when passed in to the function.

Normally this would never happen, but I had a case where I was manually passing a list of variables, and one happened to be a LinkVariable, and it would throw an error when the block was `None`.

PR #336, opened 2019-02-05
Fix error message when dma mapping fails & buffer mapping > 4GB

slacrherbst:
This PR properly detects when a virtual mapping limit has been reached and displays a helpful error message indicating how to address the issue.

This also fixes an error with mapping buffers above the 4GB boundary.


slacrherbst:
Ran overnight without errors. Merging into pre-release.

PR #339, opened 2019-02-08
Add __enter__ and __exit__ to locks

slacrherbst:
This will allow the frame and transaction locks to be used in python with blocks:
   with frame.lock():
      access frame data

### DIFF
--- a/include/rogue/interfaces/memory/TransactionLock.h
+++ b/include/rogue/interfaces/memory/TransactionLock.h
@@ -55,6 +55,11 @@ namespace rogue {
                //! lock
                void unlock();
 
+               //! Enter method for python, do nothing
+               void enter();
+
+               //! Exit method for python, do nothing
+               void exit(void *, void *, void *);
          };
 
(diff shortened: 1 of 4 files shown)


bengineerd:
I'm confused. Doesn't `enter()` need to acquire the lock, and `exit()` release it?

slacrherbst:
The way I designed the lock object is that the lock is acquired on creation and released when the object is deleted. You may also manually release and acquire the lock with method calls. In the C++ routines I typically grab the lock in a function and it auto releases when it goes out of context. I wanted to be able to do the same with python, but the with call requires the __enter__ and __exit__ routines to exist. In this case I don't need them to do anything since the lock will automatically be acquired and then be released when they are destroyed. I put some debug in to verify this is he case. I do need to make sure I am releasing the GIL as the lock is acquired. I will check that.

slacrherbst:
It turns out I was releasing the GIL properly in the transaction lock but not the frame lock.

PR #351, opened 2019-02-21
Check if LD_LIBRARY_PATH is not defined or empty

jesusvasquez333:
If `LD_LIBRARY_PATH` is empty or not defined `string(REPLACE ":" ";" HINT_PATHS $ENV{LD_LIBRARY_PATH})` will fail. So, check it before using `REPLACE`


jesusvasquez333:
@slacrherbst this was causing the docker build to fail

PR #355, opened 2019-02-26
Memory Tcp Client/Server address logging fix

slacrherbst:


bengineerd:
Should we do this globally and not just in memory.Tcp?

slacrherbst:
yes as a goal but not worth taking the time now.

PR #356, opened 2019-02-27
Eliminate boost::dynamic_bitset for PRBS computation

slacrherbst:
This change eliminates the need for the boost:: dynamic_bitset when doing the PRBS computation.

I tested sw-sw transmission against the previous version of Rogue.

### DIFF
--- a/src/rogue/utilities/Prbs.cpp
+++ b/src/rogue/utilities/Prbs.cpp
@@ -346,7 +346,7 @@ void ru::Prbs::genFrame (uint32_t size) {
    if ( genPl_ ) {
 
       // Init data
-      memcpy(data,frSeq,byteWidth);
+      memcpy(data,frSeq,byteWidth_);
 
       // Generate payload
       while ( frIter != frEnd ) {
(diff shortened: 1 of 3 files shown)


src/rogue/utilities/Prbs.cpp
```diff
@@ -170,13 +170,25 @@ void ru::Prbs::sendCount(bool state) {
...
-   for (uint32_t x=0; x < tapCnt_; x++) bit ^= data[taps_[x]];
+void ru::Prbs::flfsr(uint8_t * data) {
+   uint32_t iByte;
+   uint32_t iBit;
+   uint32_t x;
+   bool msbOut;
+   bool lsbIn = false;
+
+   for (x=0; x < tapCnt_; x++) {
+      iByte = taps_[x] / 8;
```
ruck314: instead of `iByte = taps_[x] / 8;`, wouldn't a `iByte = taps_[x] >> 3;` run faster?  That's my experience with microcontrollers but a modern CPU probably does it in a single cycle

PR #357, opened 2019-02-27
Move to standalone library for CRC

slacrherbst:
This PR eliminates the use of the boost CRC library and uses a third party open source module instead. The implementation of this CRC uses a look up table which is a static entity created at startup. The table is no longer created dynamically on each iteration.

This change has been tested against an old version of Rogue using the boost library.

### DIFF
--- a/include/rogue/protocols/packetizer/CRC.h
+++ b/include/rogue/protocols/packetizer/CRC.h
@@ -1,0 +1,1719 @@
+/**
+    @file CRC.h
+    @author Daniel Bahr
+    @version 0.2.0.6
+    @copyright
+    @parblock
+        CRC++
+        Copyright (c) 2016, Daniel Bahr
+        All rights reserved.
+
+        Redistribution and use in source and binary forms, with or without
+        modification, are permitted provided that the following conditions are met:
+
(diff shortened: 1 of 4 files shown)


ruck314:
It's not clear to me how the CRC gets intialized to 0xFFFFFFFF on the first segment sequence and how a non-first segment feedback the on the previous CRC to reintialize the CRC.  Can you explain how you are now doing this?  Also, have you've tested this with large frame PRBS frame from FW?

slacrherbst:
See these two lines:

      // Compute CRC
      if ( tmpSof ) crc_[tmpDest] = CRC::Calculate(data, size-4, crcTable_);
      else crc_[tmpDest] = CRC::Calculate(data, size-4, crcTable_, crc_[tmpDest]);

If the frame is the first in a sequence it starts from scratch and ignores the current crc value (crc_[tmpDest]). For all other frames it continues the CRC from the previous value stored in crc_[tmpDest]. 

I tested this by generating large frames in both directions between an old version of Rogue and the new version. I even purposely corrupted the CRC between frames just to make sure the multi-frame CRC was actually working.

PR #360, opened 2019-02-28
Update TCP bridge operating modes

slacrherbst:
The stream bridge will no longer queue frames when the connection is lost. It will still block until the connection is restored, but multiple frames will not be buffered.

The memory client will no longer block once the buffer is full and will return immediately when the connection is down.

### DIFF
--- a/src/rogue/interfaces/memory/TcpClient.cpp
+++ b/src/rogue/interfaces/memory/TcpClient.cpp
@@ -43,7 +43,7 @@ rim::TcpClientPtr rim::TcpClient::create (std::string addr, uint16_t port) {
 
 //! Creator
 rim::TcpClient::TcpClient (std::string addr, uint16_t port) : rim::Slave(4,0xFFFFFFFF) {
-   int32_t to;
+   int32_t opt;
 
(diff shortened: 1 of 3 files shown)


bengineerd:
The stream bridge blocking when connection goes down still might be problematic for HPS. A lot of SRP memory transactions end up going over the stream bridge to the FEBs. A blocked stream would probably freeze things up.

I don't know what the solution is though. We don't really have a mechanism to report and handle stream pipeline failure. Perhaps we need a way for a Slave to reject a Frame in its `acceptFrame` method.

slacrherbst:
Normally we just drop frames on a floor in acceptFrame when things get logged jammed. I think the only mechanism I have left is to timeout on a send and flush frames after a 1 second timeout or some other configurable timeout, just like we have for axiStream and PGP. I will look into that.

slacrherbst:
I just looked at the code. The AxiStream and PGP cards just make noise once a second but they stay locked up. We may want to consider changing that.

slacrherbst:
Most of our other interfaces drop when the connection is lost. pgp hardware does this, rssi does this, the problem is there is not a good way to know if the zeromq connection is down. At least no that I can see I will keep looking into it.

PR #364, opened 2019-03-13
Add begin and end iterators to Batcher Data and Core. Adds container reservation.

slacrherbst:
Changes Data::iterator() to Data::begin() and adds Data::end().

Changes CoreV1::header to CoreV1::beginheader() and adds CoreV1::endHeader()
Changes CoreV1::tail to CoreV1::beginTail() and adds CoreV1::endTail()

This changes also ensure that returned iterators are copies, not references.

The CoreV1 class also now has an initSize() call which reserves memory for the internal vectors used to track the records.


russell-slac:
Looks okay

PR #365, opened 2019-03-20
Allow blocks to overlap Devices in some cases

slacrherbst:
This will allow a block to overlap a device if the block is part of the Device and all of the variables in the block have the overlapEn flag set.


slacrherbst:
I think the key is that this should be painful because it is not something that normally should be done except for special cases. It really does need to be drive by each remote variable.

slacrherbst:
How many of these types of files do we have? If you point me to them I will write a script to address this.

PR #366, opened 2019-03-21
Add library path helper function

slacrherbst:
Provides a new function: pyrogue.addLibraryPath() which takes a string or list of strings containing paths which are to be added to the python library path. The passed strings can either be relative starting with '../' or absolute starting with '/'.

### DIFF
--- a/python/pyrogue/__init__.py
+++ b/python/pyrogue/__init__.py
@@ -8,6 +8,7 @@
 # contained in the LICENSE.txt file.
 #-----------------------------------------------------------------------------
 import sys
+import os
 
 MIN_PYTHON = (3,6)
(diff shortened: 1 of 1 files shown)


ruck314:
https://www.riverbankcomputing.com/static/Docs/PyQt4/qcoreapplication.html#addLibraryPath
`addLibraryPath` is probably going have a namespace conflict (like it does with PyQT4).  Maybe a more unique name?

slacrherbst:
Namespace conflicts should not be an issue since python is heavily namespaced. Global imports that ignore namespace is a bad idea. I think most code in pyrogue has been converted to avoid this but the gui portion still needs to be fixed.

PR #369, opened 2019-03-29
Update To Reset Behavior, Change SoftReset to Initialize

slacrherbst:
This PR changes the behavior of the Rogue reset commands and how they are used.

SoftReset now becomes Initialize

The three reset functions, HardReset, CountReset and Initialize are no longer recursed from the top level. Instead the default function in a device will recurse on the child devices. If the user re-implemented the reset function, they must call the super() reset function in order to properly recurse if desired. 

A new variable in root, InitAfterConfig, defines if the Initialize() method should be called after a yaml file is loaded, or after setYaml() is called directly. 

This PR also sets all blocks to stale as a result of HardReset and at startup..

ReadConfig is changed to LoadConfig, WriteConfig is changed to SaveConfig and WriteState is changed to SaveState.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -172,7 +172,8 @@ jobs:
 
       script:
         # Build conda package
-        - conda build -q conda-recipe --output-folder bld-dir -c defaults -c conda-forge -c paulscherrerinstitute
+        - conda build --debug conda-recipe --output-folder bld-dir -c defaults -c conda-forge -c paulscherrerinstitute
+        #- conda build -q conda-recipe --output-folder bld-dir -c defaults -c conda-forge -c paulscherrerinstitute
 
       after_success:
         # Upload conda package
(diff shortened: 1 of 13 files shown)


python/pyrogue/_Block.py
```diff
@@ -330,6 +330,9 @@ def __init__(self, *, offset, size, variables):
         # Set exclusive flag
         self._overlapEn = all (excMask[i] == 0 for i in range(size))
 
+        # Force block to be stale at startup
+        self._forceStale()
```
slacrherbst: Please not this change.

.travis.yml
```diff
@@ -172,7 +172,8 @@ jobs:
 
       script:
         # Build conda package
-        - conda build -q conda-recipe --output-folder bld-dir -c defaults -c conda-forge -c paulscherrerinstitute
+        - conda build --debug conda-recipe --output-folder bld-dir -c defaults -c conda-forge -c paulscherrerinstitute
```
ruck314: Is the `--debug` flag the new default?  Or was this an accidental comment?

slacrherbst: Sneaking in a debug flag to figure out why the mac conda build is failing.

bengineerd:
This PR should also change readConfig and writeConfig to loadConfig and saveConfig.

slacrherbst:
Ok. I changed those functions. They were already named that in the GUI.

PR #370, opened 2019-03-29
Increase arg/response string size

slacrherbst:
Increase arg and response string sizes to 10MB.


slacrherbst:
I added some additional changes and support for commands that will return values. The command change will be in another PR.

PR #375, opened 2019-04-17
HPS and KPIX Development Bug Fixes and Enhancements

bengineerd:
There are several bug fixes and enhancements that have piled up when developing rogue for HPS and KPIX.

### Bug fixes
 - `BaseBlock` now implements `_forceStale()`, since this is now called on all blocks upon startup
 - `keyValueUpdate()` in _Root.py now properly pushes new empty dicts at keys that didn't previously exits
 - `BaseVariable` gets a default `offset=0`
 - `SideBandSim` had the push and pull ports switched
 - Fix logging message in TcpCore.cpp

### Enhancements
 - Added built in commands for setting a RemoteCommand with argument (`setWithArg`) and setting and verifying with arg (`setAndVerifyArg`)
 - There is a new `ArrayDevice` class for creating a Device that contains an array of identical `Devices` 
 or `Variables`
 - `RemoteVariable` implements `_setDefault()` so that defaults from the parent Device can be set on startup
 - `StreamWriter` now sets the `runCount` back to 0 when a new file is opened
 - Prbs Devices can now be instantiated with a stream to connect to

Note: not quite ready to actually merge yet, but wanted to get the process started.

#### ArrayDevice Examples
These examples demonstrate complex array creation where each array member gets different constructor arguments. `arrayArgs` can also be a single `dict` that is then passed to each array member, with each member constructor receiving the same arguments. That is the primary intended use case, where you just want to wrap a group of nearly identical Devices in a parent Device. Anything more complex than these examples and you should probably just declare a new class manually.

```python
# Create a device with a LinkVariable pointing to each link status
self.add(pr.ArrayDevice(
    name='FebLinkStatus',
    arrayClass=pr.LinkVariable,
    number=12,
    arrayArgs= [
        {'name': 'Link',
         'mode': 'RO',
         'variable' : self.PgpStatus.Pgp2bAxi[hps.constants.FEB_LINK_MAP[i]].RxRemLinkReady}
        for i in range(12)]))

# Create an ArrayDevice containing 10 FebFpga Devices
# Each FebFpga can receive different arguments
# febMemBases is a list of MemorySlave objects
self.add(pr.ArrayDevice(
    name='FebArray',
    arrayClass=hps.FebFpga,
    number=len(febMemBases),
    arrayArgs = [
        {'number' : i,
         'memBase': dest,
         'expand': False,
         'enableDeps': [self.PgpStatus.Pgp2bAxi[hps.constants.FEB_LINK_MAP[i]].RxRemLinkReady]}
        for i, dest in enumerate(febMemBases)]))
```


python/pyrogue/_Variable.py
```diff
@@ -447,6 +449,10 @@ def verify(self):
     def base(self):
         return self._base
 
+    def _setDefault(self):
```
slacrherbst: This is to avoid the device functions called with normal setDisp?

python/pyrogue/_Root.py
```diff
@@ -696,9 +695,12 @@ def keyValueUpdate(old, key, value):
     d = old
     parts = key.split('.')
     for part in parts[:-1]:
-        d = d.get(part, {})
+        if part not in d:
```
slacrherbst: Does the default return not work properly?

python/pyrogue/_Device.py
```diff
@@ -526,6 +527,28 @@ def _decorator(func):
             return func
         return _decorator
 
+class ArrayDevice(Device):
```
slacrherbst: Can you drop an example in the doc section, maybe in a future notes directory.

python/pyrogue/utilities/prbs.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, *, name, width=None, taps=None, **kwargs ):
         if taps is not None:
             self._prbs.setTaps(taps)
 
+        if stream is not None:
```
slacrherbst: Why?

python/pyrogue/_Root.py
bengineerd: It needs to return a new dict, but also push that new dict into `d` at `part` before assigning over `d`. `d.get(part, {})` doesn't do the push.
What i was seeing was that if `keyValueUpdate()` was called with an empty top level dict, it would not actually fill it with anything.

python/pyrogue/_Device.py
bengineerd: I added some examples in the PR description. I'll do it in the documentation once we merge it.
I'm open to discussion on this implementation. I wanted a quick way to group a bunch of nearly identical Nodes into a parent, but maybe this makes the code harder read and follow.

python/pyrogue/utilities/prbs.py
bengineerd: We can probably remove this whole feature. The thought was that a Prbs Device is sort of useless without a stream connected to it, so you might as well have a way to pass it one right when you make it. I don't actually use it anywhere in HPS or KPIX though.
I also made `PrbsPair` but then never ended up using it anywhere.

python/pyrogue/_Variable.py
bengineerd: I need to go back and trace why it is, but my recollection is that the `defaults` passed to a Device constructor don't work unless this is here.

bengineerd: I'm still trying to trace this, but I think it's that `Block._setDefault()` needs to be called to properly stage defaults for RemoteVariables. `BaseVariable.setDefault()` will just call `self.setDisp(write=False)`, which doesn't quite do the right thing.

It is possible though that some underlying `Block.set()` implementation has since changed and that this is no longer necessary.

slacrherbst:
What is the status of this. We may need to push a release soon, what of these changes do we want to keep and what do we want to adjust.

bengineerd:
I'm ok with the whole thing getting merged.

PR #377, opened 2019-04-23
 v3.4.1 release candidate

ruck314:
### Description
- Add ability to return data from command, use command default, add new root commands #371
- Command interface fixes #372
- Add path check to addLibrary #374
- Add path alias to Block, print full path in overlaps #373
- Update To Reset Behavior, Change SoftReset to Initialize #369
- Increase arg/response string size #370


slacrherbst:
I will generate a release as soon as the pending PRs are approved.

PR #383, opened 2019-05-03
Add ZMQ API For Future Clients

slacrherbst:
Creates a new ZMQ server interface for interacting with the root tree. The ZMQ server creates two ports. The first is a pub/sub port which allows clients to monitor variables updates in the rogue tree. The same yaml string that is send to the data file is sent over this pub/sub interface.

The second interface is a request/response interface. This allows clients to interact with the rogue tree. The client will send a dictionary which includes the path to the note it wants to interact with, the attribute name it wants to call, a *args list of args and a **kwargs list of keyword args. A yaml string of the return value(s) is returned. 

This interface should be flexible enough to interact with all rogue attributes.


slacrherbst:
Closing because this is include in #385

PR #385, opened 2019-05-12
Add remote zmq client to replace Pyro4 library

slacrherbst:
To enable the zmq server:

root.start(zmqPort=xxxx)

Two ports will be used, the passed port and the one above the passed port. If 9000 is passed, ports 9000 and 9001 will be used. 

To start the zmq client:

client = pyrogue.VIrtualClient(addr="localhost",port=xxxx)

client.root is the root of the remote pyrogue instance and can be used in scripts or passed to the GUI.

This PR also makes ZMQ and BZIP mandatory libraries and removed the ApiWrapper and shared memory control interfaces.


slacrherbst:
We also need to decide if the zmq server should be enabled by default.

slacrherbst:
Ok. This is working well with HPS. Ready for official PR.

PR #386, opened 2019-05-14
Add GUI pop up menu to enable read/write methods on variables on devices

slacrherbst:
Right clicking on any variable box will provide a popup menu which will enable the user to read or write the specific variable or the entire containing device. Device read and writes are not recursive. The read variable and write variable options are dependent on the variable mode.

### DIFF
--- a/python/pyrogue/_Device.py
+++ b/python/pyrogue/_Device.py
@@ -159,12 +159,12 @@ class Device(pr.Node,rim.Hub):
         # Variable interface to enable flag
         self.add(EnableVariable(enabled=enabled, deps=enableDeps))
 
-        self.add(pr.LocalCommand(name='ReadDevice', value='', hidden=True,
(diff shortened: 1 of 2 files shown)


bengineerd:
When you use the menu to read a LinkVariable, what should happen?

Currently it will just call the variable `get()` function, which usually just grabs the `value()` from any dependent variables. So it doesn't do a read on the underly dependencies.

Maybe that's what we want, but consider how the poller works. When you pass a LinkVariable to the poller, it traces all the blocks in all the dependencies and adds them to the poll list. Maybe the GUI menu read should do something similar? It seems like when you call `read` in the GUI menu, you'd want it to do the read on the underlying dependencies.

Also, maybe there could be menu items for `Read Device` and `Read Device Recursive`? I'm finding that I want to be able to make the recursive call when debugging.

slacrherbst:
I added the recursive options.

PR #387, opened 2019-05-14
Remove enable changed feature

slacrherbst:
Places where this was used should be replaced with linkVariable which intercept the top level call and generate transactions as necessary. Right now I think HPS is the only place this is used.


slacrherbst:
We will keep this feature.

PR #398, opened 2019-06-03
Define Types for RSSI GUI pages

slacrherbst:


slacrherbst:
Bad commit. Deleting.

PR #400, opened 2019-06-05
Make pickle operation recursive, Use JSON for ZMQ Interface

slacrherbst:
This change makes the node pickle operation recursive, allowing the full tree to be picked in a single operation.

Also addresses a bug where some sub-nodes were not being setup properly and causing errors when right clicking "ReadDevice" on the GUI.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -39,6 +39,7 @@ requirements:
      - epics-base=3.14.12.8
      - zeromq
      - pyyaml
+     - jsonpickle
      - pyzmq
      - parse
      - click
--- a/conda.yml
+++ b/conda.yml
@@ -12,6 +12,7 @@ dependencies:
   - ipython
   - cmake
   - pyyaml
+  - jsonpickle
   - pyzmq
   - parse
   - click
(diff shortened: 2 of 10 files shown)


slacrherbst:
This PR is ready.

bengineerd:
When I toggle the `enable` variable on a Device in the GUI, I see :
```
  File "/u/re/bareese/anaconda3/envs/kpix-env/lib/python3.7/site-packages/pyrogue/_Virtual.py", line 228, in _doUpdate
    n._doUpdate(val)
  File "/u/re/bareese/anaconda3/envs/kpix-env/lib/python3.7/site-packages/pyrogue/_Virtual.py", line 166, in _doUpdate
    func.varListener(self.path,val.value,val.valueDisp)
AttributeError: 'NoneType' object has no attribute 'value'
```

slacrherbst:
Oh wow. I think this exposed a bug with Enable

slacrherbst:
Yes. See this:
https://github.com/slaclab/rogue/pull/400/commits/e087e4c7ad9a1d1ff88f2e84f9a900e75141319a

PR #402, opened 2019-06-07
Support 64-bit file sizes

slacrherbst:


slacrherbst:
Interrogating the C return values won't work, mostly because a lot of the calls are wrapped in lambdas.

PR #404, opened 2019-06-07
Allow user to specify typeStr for LocalVariables

slacrherbst:
This is required to properly map variable type to external clients.


slacrherbst:
Bad PR

PR #413, opened 2019-06-18
Add generic data reader class

slacrherbst:
Allows for non streaming read of Rogue data files. Two new classes are created:

pyrogue.utilties.fileio.FileReader
pyrogue.utilties.fileio.FileData

FileReader.next reads the next frame from the file returning a structure containing header and data. A config channel can be passed and the configuration will be cached in the FileReader structure for access.

To read a file:

fread = pyrogue.utilities.fileio.Filereader('data.dat',configChan=100)
head = fread.next
np_data = numpy.frombuffer(head.data,numpy.uint32)

### DIFF
--- a/python/pyrogue/utilities/fileio.py
+++ b/python/pyrogue/utilities/fileio.py
@@ -132,40 +132,45 @@ class FileHeader(object):
 
 class FileReader(object):
 
-    def __init__(self, filename, configChan=None, dtype={}):
+    def __init__(self, filename, configChan=None):
         self._filename   = filename
-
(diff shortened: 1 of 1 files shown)


bengineerd:
I've been wanting something like this that allows next() to be advanced manually.

Should this be in C++ though?

slacrherbst:
The only reason to make it C++ would be to allow it to be used in C++ software. I figured we may add the ability to skip around the data by added new calls such as:

fr.next(skip=3,chan=1)

or something like that.

PR #417, opened 2019-06-26
Features and Enhancements from HPS development

bengineerd:
This PR is not ready to merge. It exists so we can keep a handle on the HPS divergence from mainline rogue.

 - Add `Device.setPollInterval()` to change the polling interval of several variables at once
 - Remove calls to unimplemented `Model.mask` function
 - `MemoryDevice` now actually works. 
       - Needs some work to make it thread safe over an entire memory load
 - The `root` updateThread has exception handling logic so it doesn't crash
 - Increase the `AxiStreamDma` `runThread` timeout

### DIFF
--- a/python/pyrogue/_Memory.py
+++ b/python/pyrogue/_Memory.py
@@ -46,6 +46,26 @@ class MemoryDevice(pr.Device):
     def _buildBlocks(self):
         pass
 
+    @pr.expose
+    def set(self, offset, values, write=False):
+        with self._memLock:
+            self._setValues[offset] = values
+            if write:
+                self.writeBlocks()
+                self.verifyBlocks()
+                self.checkBlocks()
+
+    @pr.expose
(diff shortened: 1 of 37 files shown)


python/pyrogue/_Memory.py
```diff
@@ -46,6 +46,26 @@ def __init__(self, *,
...
+    def set(self, offset, values, write=False):
+        with self._memLock:
+            self._setValues[offset] = values
+            if write:
+                self.writeBlocks()
+                self.verifyBlocks()
+                self.checkBlocks()
+
+    @pr.expose
+    def get(self, offset, numWords):
```
slacrherbst: Should this have a read=True attribute to just return the value without a read?

bengineerd:
This wasn't quite ready to merge. The MemoryDevice was still a bit messy. Will probably be ok though.

PR #423, opened 2019-07-10
Change default of `expand` from True to False

ruck314:
### Description
- Change default of `expand` from True to False


bengineerd:
This is already merged, but I think the Root needs to expand by default. Otherwise the GUI opens fully minimized every time. Devices under the root can minimize by default.

ruck314:
I agree.  Refer to https://github.com/slaclab/rogue/pull/432

PR #427, opened 2019-07-19
Build fix for EPICS 7 which moves pcas to a separate package.

bhill-slac:
EPICS 7 moves pcas, the portable CA server, to a separate EPICS style module.
i.e. paths are: ${EPICS_PCAS_ROOT}/include and ${EPICS_PCAS_ROOT}/lib/${EPICS_HOST_ARCH}

This fix could use some cleanup as it leaves  a number of variable names w/ V3 intact,
but I wanted to minimize changed lines.

If you want I can submit another pull request that cleans up the V3 vs V7 stuff and the
cookie-cutter apple vs linux  support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,24 +136,48 @@ if((NOT NO_PYTHON) AND (NOT NO_EPICS) AND DEFINED ENV{EPICS_BASE})
 
    if(APPLE)
 
-      set(EPICSV3_INCLUDES  ${EPICSV3_BASE_DIR}/include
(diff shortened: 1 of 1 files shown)


slacrherbst:
EPICS_PCAS_ROOT is passed on the cmake command line? Is there a way to pick this up in the environment?

bhill-slac:
No.  PCAS isn't normally set in env.

On 7/19/2019 10:42 AM, Ryan Herbst wrote:

EPICS_PCAS_ROOT is passed on the cmake command line? Is there a way to pick this up in the environment?

PR #428, opened 2019-07-23
Pass VariableValue class to update functions

slacrherbst:
This PR changes the API for the variable update functions. The API changes from:

varUpdated(path, value, disp)

to 

varUpdated(path,varValue)

This will impact any application code which uses these callbacks. This will enable more information, such as alarm state, to be passed as part of the update. This also removes the check for varUpdated as part of the passed class. You must now either pass a variable or the function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,7 +99,7 @@ jobs:
       script:
         # Run tests with coverage
         - cd $TRAVIS_BUILD_DIR
-        - python3 -m pytest --cov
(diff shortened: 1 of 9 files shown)


bengineerd:
I merged this branch into my kpix-dev branch and refactored my update functions. It seems to work.

slacrherbst:
Its still failing in travis. Trying to figure that out now.

PR #431, opened 2019-07-29
v3.6.1 release candidate

ruck314:
### Description
- Set root's default to zmqPort=None
- Change default of `expand` from True to False #423
- Fix epics enum interface #424
- Poller Enhancements #425
- Build fix for EPICS 7 which moves pcas to a separate package. #427
- Fix build error deriving zeromq location from LD_LIBRARY_PATH. #426


slacrherbst:
Waiting for other updates.

PR #434, opened 2019-08-01
Fix config load issue

slacrherbst:
This fixes an issue where a memory block bit is first set to 1 using a configuration file wildcard and then later set to zero. The current code would not clear the bit.

### DIFF
--- a/python/pyrogue/_Block.py
+++ b/python/pyrogue/_Block.py
@@ -391,7 +391,7 @@ class RemoteBlock(BaseBlock, rim.Master):
             srcBit = 0
             for x in range(len(var.bitOffset)):
                 self._copyBits(self._sData, var.bitOffset[x], ba, srcBit, var.bitSize[x])
-                self._setBits( self._sDataMask, var.bitOffset[x], var.bitSize[x])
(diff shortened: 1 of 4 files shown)


ruck314:
https://github.com/slaclab/atlas-altiroc-daq/blob/v2.1.2/software/python/common/_AltirocSlowControl.py#L27-L39
I have confirmed this fixes the issue in for both methods:
Method#1:
```
        def addReg(name,description,bitSize,bitOffset,value,base):
            
            remap = divmod((bitOffset-1),32)
        
            self.add(pr.RemoteVariable(  
                name        = name, 
                description = description,
                base        = base,
                offset      = (remap[0]<<2),
                mode        = 'RW', 
                bitSize     = bitSize, 
                bitOffset   = remap[1],
                # value       = value, 
            ))
```
Method#2
```
        def addReg(name,description,bitSize,bitOffset,value,base):
                        
            self.add(pr.RemoteVariable(  
                name        = name, 
                description = description,
                base        = base,
                offset      = 0x000,
                mode        = 'RW', 
                bitSize     = bitSize, 
                bitOffset   = bitOffset-1,
                # value       = value, 
            ))     

```

PR #436, opened 2019-08-01
Remove group from gui

slacrherbst:
This removes the concept of a "group" from the gui. Instead the roots are added right at the top level of the tree, removing an unnecessary level of indent.


ruck314:
https://github.com/slaclab/dev-board-examples/blob/master/software/rogue/scripts/DevBoardGui.py#L223

ruck314:
This appears that it would break all the GUIs.  Shouldn't we keep the agrument at the top level and adding a "we are planning to depreciate this message in a later release message"?  Else it will break current builds without give people time to migrate.

slacrherbst:
Ok I added the warning.

PR #437, opened 2019-08-01
Call enable variable update chain directly

slacrherbst:
This will cause the update chain for the Enable variable to be caused directly in response to set() instead of being pushed to the update queue.


slacrherbst:
I fixed that bug. self.__listeners was private.

slacrherbst:
Should we merge this pr or leave the enable updates in the update thread?

PR #439, opened 2019-08-05
v3.6.1 release candidate

ruck314:
### Description
- Set root's default to zmqPort=None
- Change default of `expand` from True to False #423
- Fix epics enum interface #424
- Poller Enhancements #425
- Build fix for EPICS 7 which moves pcas to a separate package. #427
- Fix build error deriving zeromq location from LD_LIBRARY_PATH. #426
- Build fix for EPICS 7 which moves pcas to a separate package. #427
- Root needs to expand by default #432
- Expose Root.waitOnUpdate() on ZMQ Server #433
- Change combo box display #429
- Remove group from gui #436
- Fix config load issue #434
- Pass VariableValue class to update functions #428
- Fix missing bit overlap check when adding variables to a block. #438

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,7 +99,7 @@ jobs:
       script:
         # Run tests with coverage
         - cd $TRAVIS_BUILD_DIR
-        - python3 -m pytest --cov
+        - python3 -m pytest -s --cov
 
       after_success:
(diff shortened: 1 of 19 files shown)


slacrherbst:
Release in progress.

PR #442, opened 2019-08-06
Add alarm levels to Variables

slacrherbst:
This PR adds an alarm state to pyrogue Variables. For each variable you can high and low alarm and warning levels. These levels will then update the alarm state for the variable which is reported by EPICS. The GUI will also indicate the alarm state with a new alarm column which replaces the previous base column.

There are a couple of items to consider. First what alarm status and state values should we define within Rogue?. Right now we have borrowed from epics and indicate the following status values:

None
AlarmLoLo
AlarmHiHi
AlarmLow
AlarmHigh

with the corresponding severities:
None
AlarmMinor
AlarmMajor

Secondly should we replace the GUI base column with alarm, add a new column, or simply indicate the alarm state by changing the color of the field?


bengineerd:
I think we should have a new column in the GUI for alarm state. That column should display a color indicator of some kind along with the severity text.

slacrherbst:
In the current PR I replace the base column with an alarm column that shows None, Minor, or Major. The box itself will show yellow when in minor alarm and red when in major alarm.

PR #443, opened 2019-08-06
Fix spin box updates

slacrherbst:
Fixes the spin box updates with the new VariableValue passing.

### DIFF
--- a/python/pyrogue/gui/variables.py
+++ b/python/pyrogue/gui/variables.py
@@ -171,7 +171,7 @@ class VariableLink(QObject):
             else:
                 self._variable.setDisp(self._widget.text())
 
-    def varListener(self, path, value):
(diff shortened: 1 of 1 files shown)


python/pyrogue/gui/variables.py
```diff
@@ -187,7 +187,7 @@ def varListener(self, path, value):
                     self.updateGui.emit(i)
             elif isinstance(self._widget, QSpinBox):
                 if self._widget.value != value.value:
-                    self.updateGui.emit(value)
+                    self.updateGui.emit(value.value)
```
ruck314: `value.value` is a little strange.  Maybe the `def varListener(self, path, value):` should be changed to `def varListener(self, path, var):` then it would be `var.value`, which is more intuitive when reading the code

PR #444, opened 2019-08-06
Replace hidden flag with visibility field

slacrherbst:
This PR replaces the hidden flag with a visibility field. The higher the visibility level the more likely it will be visible. This field is a simple int with a proposed scale of 0 (hidden) to 100 (always visible).

By default new nodes are created with a visibility value of 25. The root has a default visibility of 100 and Devices have a default visibility value of 75. Important variables such as the root level timestamp and the device enable variables have a default visibility of 50.

Visibility is inherited in that if the parent has a lower visibility than the child, the lower of the two will be effective. 

Both the GUI top level and the Epics top level take a minVisibility flag. Of course the epics visibility can be overridden by passing a map as before.

Attempts have been made to keep this backward compatible with a deprecation warning. We should remove Hidden=False from all existing code immediately and transition the remaining when the new version of Rogue gets adopted more globally. We may have to remove the deprecation warning for a while.

I will continue to tweak this over the next day, but I wanted to get any feedback.


ruck314:
@slacrherbst: To help with the transition from `hidden` to `visibility`, I think that we need to keep the 'hidden' arg in all the functions for the next few tag release cycles but default hidden to "None." So if hiddent is not None, then you could add a warning message about how we are going to depreciate 'hidden' completely in future release and replace it with `visibility`.

slacrherbst:
The current code keeps hidden but generates a warning message. We may want the morning message suppressed in the short term though....

bengineerd:
The 0-100 range feels a bit too wide. Probably 0-10 is plenty.

slacrherbst:
I figure that if we have 0 - 100 we can put the standard steps at large values 0, 25, 50, 75, 100 and allow room for someone to find an in between stop for application specific visibilities. If you go with 0 - 10 the standard visibilities would end up at 0, 2,5,7,10 and you may run out of room particularly on the low end side.

PR #445, opened 2019-08-07
Add save enable flag to variables

slacrherbst:
This adds the saveEn attribute to variables with a default value of True. If saveEn is set to false the variable will not be included in the configuration save. It will still be included in state dumps and dumps to the data file. This only applies to the SaveConfig and GetYamlConfig commands.

### DIFF
--- a/python/pyrogue/_Root.py
+++ b/python/pyrogue/_Root.py
@@ -147,10 +147,16 @@ class Root(rogue.interfaces.stream.Master,pr.Device):
         self.add(pr.LocalCommand(name='SetYamlConfig', value='', function=lambda arg: self._setYaml(arg,False,['RW','WO']), hidden=True,
                                  description='Set configuration from passed YAML string'))
 
-        self.add(pr.LocalCommand(name='GetYamlConfig', value=True, function=lambda arg: self._getYaml(arg,['RW','WO']), hidden=True,
+        self.add(pr.LocalCommand(name='GetYamlConfig', 
+                                 value=True, 
+                                 function=lambda arg: self._getYaml(readFirst=arg,modes=['RW','WO'],saveEnFilt=True), 
+                                 hidden=True,
                                  description='Get current configuration as YAML string. Pass read first arg.'))
 
-        self.add(pr.LocalCommand(name='GetYamlState', value=True, function=lambda arg: self._getYaml(arg,['RW','RO','WO']), hidden=True,
+        self.add(pr.LocalCommand(name='GetYamlState', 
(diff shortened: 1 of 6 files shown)


python/pyrogue/_Root.py
```diff
@@ -379,14 +385,14 @@ def _sendYamlFrame(self,yml):
         frame.write(b,0)
         self._sendFrame(frame)
 
-    def _streamYaml(self,modes=['RW','RO']):
+    def _streamYaml(self,modes=['RW','RO','WO'],saveEnFilt=False):
```
ruck314: I dont' think 'WO' should be included here

slacrherbst: Before it was not. "WO" variables are part of the configuration and should be included in the stream dump since they represent the current config.

slacrherbst:
To go along with this change. We may also need to consider a rexexp filter parameter to create config files which dump a sub-set of fields, or maybe start the save at a particular branch of the device tree....

slacrherbst:
But that would come in a different PR

bengineerd:
I know we went back and forth a bit before deciding on how to do this, but now that it's coded it feels somewhat kludgy. I also don't think creating new `modes` is the right thing either. The `mode` specifies accessibility, and adding "saveability" to that would conflate two separate things.

Perhaps variables need a "type" property (or some better name) that would basically encompass the idea of "Configuration" and "Status" Variables that we had in the old system, with the ability to add new types and to filter tree dumps based upon the type. Writable variables would get a default type (name TBD but probably "configuration") that is included in both state and config dumps. Users could then override the default type as needed to include or exclude variables from state and or config dumps.

slacrherbst:
Another option would to make this field more expansive, or maybe add change your type field to a category field. Possible options would be:
category = 'Configuration' (default)
category='Linked' (for ones that don't see to be saved)?

And then you could add categories for things you may want to save separately later into separate configuration files.

bengineerd:
Yes, but then perhaps `category` could be a list? Basically a list of tree crawls that it will be included in.

python/pyrogue/_Root.py
slacrherbst: Plus this flag will allow the user to specifically set the variable to not be included if it should not.

slacrherbst:
Maybe we screwed up visibility then. Maybe these need to be combined into a single field which defines categories and we can filter the gui, epics and configuration dumps based upon categories.

slacrherbst:
Moving to a groups based approach.

PR #446, opened 2019-08-07
Add process device

slacrherbst:
This PR creates a base class Process which is a sub-class of device. This Process class is meant to host a long running operation. The base class has a start and stop command as well as a running flag and percent complete status variable. The user can add configuration and output variables as required.


.gitignore
```diff
@@ -6,6 +6,7 @@ setup_env.csh
 setup_env.sh
 *.swp
 *.orig
+*.rstx
```
ruck314: Out of curiosity, what's the reason for the *.rstx? I am not familiar with this file type.

slacrherbst: It is part of the documentation. It appears to be an output of the .rst compile.

bengineerd:
Should we make LoadConfig() a Process? It might be overkill for some applications, but for others like HPS it would be useful.

slacrherbst:
Yes that may make sense for anything that might take longer than a standard process. Initialize as well maybe. Basically anything you want to start and monitor until it finishes.

bengineerd:
I think RunControl should go in a _RunControl.py file. One motivation for doing file names the way we do is to make class definitions easier to find, so it's a bit confusing for it to be inside of _Process.py when `RunControl` doesn't even inherit from `Process`.

PR #447, opened 2019-08-13
Add ability to define command return type

slacrherbst:
This adds a retType= arg to command creation in order to identify the return type value for commands which return values. This adds the accompanying property to the command:

self.retTypeStr

which returns None for commands that do no return a value.


bengineerd:
I assume there's some integration that needs this? What is it?

slacrherbst:
Its and epics4 thing but it may also be useful for other processing.

PR #464, opened 2019-08-26
Allow process device to be called with an arg

slacrherbst:
This adds the call() and __call__ methods to the Process device to allow it to be treated as a command. The creator now includes the argVariable arg which indicates which sub-variable to update with the value passed in the call. This could be used to replace the "LoadConfig" Command with a Process.

LoadConfig(argVariable='FileName')

LoadConfig.Start.call("config.yaml")
LoadConfig.Start("config.yaml")
LoadConfig.call("config.yaml")
LoadConfig("config.yaml")

all can be used to start the process.

### DIFF
--- a/python/pyrogue/_Command.py
+++ b/python/pyrogue/_Command.py
@@ -82,8 +82,7 @@ class BaseCommand(pr.BaseVariable):
     def retTypeStr(self):
         return self._retTypeStr
 
-    @pr.expose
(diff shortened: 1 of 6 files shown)


slacrherbst:
I wonder if the call() interface should be blocking, waiting for the process to complete before returning? That allows a process to have both a blocking and non-blocking interface.

PR #474, opened 2019-08-30
Make some functions public so they can be access by users 

slacrherbst:
This moves some root methods from private to public so that users are able to create their own versions of LoadConfig and SaveConfig which may make use of the new groups feature.

i.e. SaveThresholds....


bengineerd:
So a while ago on HPS I tried to do the following in my HpsRoot:

```python
def _loadConfig(self, arg):
    self._pollQueue.stop()
    super()._loadConfig(arg)
    self.pollQueue._start()
```

I had some issues doing that if exceptions were thrown in `super()_.loadConfig()`, though I can't remember what it was exactly. I think it was that the exceptions would no longer be caught and logged, and would instead crash the whole thing.

slacrherbst:
My idea is not to sub-class and redefine loadConfig. The idea is the user would hide the LoadConfig command (or future process) and define their own version with a new name. LoadHpsConfig. They would then call the supporting loadConfg function with different args. Thinking about it maybe we need to rename it to something less similar. Maybe loadConfigHelper.... or something.

bengineerd:
It's be helpful to have a few usage examples.

slacrherbst:
Here is an example of what I am thinking:

self.add(pr.LocalCommand(name='SaveTHoldConfig', value='', 
                                 function=lambda arg: self.saveYaml(name=arg,
                                                                    readFirst=True,
                                                                    modes=['RW','WO'],
                                                                    incGroups='TholdConfig',
                                                                    excGroups='NoConfig',
                                                                    autoPrefix='config'),
                                 hidden=True,
                                 description='Save configuration to passed filename in YAML format'))

Looking back it it I am not sure why I was worried about name conflicts. The names of the methods are very different from the exposed commands.

slacrherbst:
Maybe we should also consider letting users update the called function of a command. So in your example:

def _myConfig(self, arg):
    self._pollQueue.stop()
    self.loadYaml(name=arg,
                          writeEach=False,
                          modes=['RW','WO'],
                          incGroups=None,
                          excGroups='NoConfig'),
    self.pollQueue._start()

root.LoadConfig.function=self._myConfig

PR #475, opened 2019-08-31
Add poll blocking context

slacrherbst:
This PR adds a poller blocking context to temporarily pause polling while the context is active:

with root.pollBlock():
     --- do some stuff

### DIFF
--- a/python/pyrogue/_PollQueue.py
+++ b/python/pyrogue/_PollQueue.py
@@ -41,11 +41,11 @@ class PollQueue(object):
         self._pq = [] # The heap queue
         self._entries = {} # {Block: Entry} mapping to look up if a block is already in the queue
         self._counter = itertools.count()
-        self._lock = threading.RLock()
-        self._update = threading.Condition()
+        self._clock = threading.Condition(threading.RLock())
         self._run = True
         self._pause = True
         self._root = root
+        self._bcount = 0
         self._pollThread = threading.Thread(target=self._poll)
 
         # Setup logging
@@ -56,7 +56,7 @@ class PollQueue(object):
         self._log.info("PollQueue Started")
 
     def _addEntry(self, block, interval):
-        with self._lock:
+        with self._clock:
             timedelta = datetime.timedelta(seconds=interval)
             # new entries are always polled first immediately 
             # (rounded up to the next second)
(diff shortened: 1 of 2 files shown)


python/pyrogue/_PollQueue.py
```diff
@@ -41,11 +41,11 @@ def __init__(self,*, root):
         self._pq = [] # The heap queue
         self._entries = {} # {Block: Entry} mapping to look up if a block is already in the queue
         self._counter = itertools.count()
-        self._lock = threading.RLock()
-        self._update = threading.Condition()
+        self._clock = threading.Condition(threading.RLock())
```
bengineerd: Maybe name this `self._cond`? It took me a sec the realize that `clock` is short for Condition Lock.

python/pyrogue/_PollQueue.py
```diff
@@ -41,11 +41,11 @@ def __init__(self,*, root):
...
         self._run = True
         self._pause = True
         self._root = root
+        self._bcount = 0
```
bengineerd: Also just name this `_blockCount`.

python/pyrogue/_PollQueue.py
slacrherbst: Yeah I can rename. The main point is that you can use one object for all locking here. maybe condLock?

python/pyrogue/_PollQueue.py
slacrherbst: Ok. will do.

slacrherbst:
Ok so the global lock you are thinking of would be for extreme cases where you want to block all other entry points (gui, epcis, zmq, etc, poll) while the operation occurs? So the blocking thread would do:

with root.exclusive():
    do some stuff

and all of the other entry points would need to do:

with root.nonexclusive():
    do some stuff.

slacrherbst:
We would need to be careful where we place the:

with root.nonexclusive():

call since it could kill performance if done wrong. But maybe we start with this model for poll blocking and we can expand upon it with careful testing.

slacrherbst:
I think we should stick with the poll block for now. Since some of the clients access variables directly it is hard to set a good entry point for other locks.

PR #478, opened 2019-09-03
Reorganize fileio directory and update FileReader

slacrherbst:
This PR breaks the fileio.py file into multiple files under a fileio sub-directory. Existing imports will still work as before.

I also have updated the currently unused FileReader to provide a path lookup mechanism for configuration as well as return the data contents as a numpy array.


slacrherbst:
This PR is ready for review.

PR #482, opened 2019-09-04
Add data receiver Device

slacrherbst:
This adds a data receiver Device which contains variables indication rx count, bytes etc. It also includes a method for users to override to re-organize the stored numpy array to match the data structure.

### DIFF
--- a/python/pyrogue/_DataReceiver.py
+++ b/python/pyrogue/_DataReceiver.py
@@ -34,15 +34,33 @@ class DataReceiver(pr.Device,ris.slave):
                                   pollInterval=1,
                                   description='Byte Rx Counter'))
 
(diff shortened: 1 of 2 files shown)


python/pyrogue/_DataReceiver.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.add(pr.LocalVariable(name='Data',       
+                                  hidden=True,
+                                  value=numpy.empty(shape=0, dtype=numpy.Int8, order='C'),
+                                  description='Data Frame Container'))
+
+
+    def _acceptFrame(self, frame):
+
+        # The following block of lines needs to be made more effecient, JJ is working on this
```
slacrherbst: This should be better. Maybe something like:

frame.getNumpy()

PR #489, opened 2019-09-10
Allow for read of configuration directory containing multiple files

slacrherbst:
This PR allows the loading of multiple configuration files when a directory is passed as an arg to LoadConfig.

Any .yml or .yaml files in the passed directory will be loaded in sorted order. Each file is parsed and applied to the management tree individually with the write() call generated after all configuration files have been loaded. 

This PR also allows the root node to be aliased with the name 'root'.

### DIFF
--- a/python/pyrogue/_Root.py
+++ b/python/pyrogue/_Root.py
@@ -528,8 +528,9 @@ class Root(rogue.interfaces.stream.Master,pr.Device):
 
         # Passed path is a directory
         if os.path.isdir(name):
-            lst = glob.glob('{}/*.yml'.format(name))
-            lst.extend(glob.glob('{}/*.yaml'.format(name)))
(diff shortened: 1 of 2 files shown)


ruck314:
Have you’ve considered changing LoadConfig from a string to a list of strings? Then you could better coordinate the order of YAML loading if you have not thought ahead about filename sorting or don’t want to load all the YAML files in a directory.

slacrherbst:
I added the ability to provide a comma separated list of files to the function:
LoadYaml("file1.yml,file2.yml")

Or a python list of files:
LoadYaml(["file1.yml","file2.yml"])

This also allows multiple files to be selected in the GUI.

slacrherbst:
You can also now included directories in the list and their contents will be inserted at that location in the list.

PR #505, opened 2019-09-18
Fix corner case where block and device have the same offset

slacrherbst:


ruck314:
Is it possible to add a test script to the TravisCI checking for these types of corner cases? It would also help to improve the code coverage as well.

PR #512, opened 2019-09-19
ESROGUE-210 - Add UART Memory Interface

bengineerd:
Provides software interface to https://github.com/slaclab/surf/blob/master/protocols/uart/rtl/UartAxiLiteMaster.vhd

Implementation of this JIRA Ticket: https://jira.slac.stanford.edu/browse/ESROGUE-210


slacrherbst:
I worry that this creates a blocking point in the main transaction thread. It might be better if the transaction was queued to a work thread which would execute the transaction and issue the done when complete.

slacrherbst:
Also this does not work with the new transaction api which seperates:
transaction.done()
and 
transaction.error("Some error string")

And the error constants are not longer defined. i.e. (ProtocolError, etc)

slacrherbst:
Ok. I updated the done() and error() calls.

slacrherbst:
See here for a possible threaded version:
https://github.com/slaclab/rogue/blob/ESROGUE-210-thread/python/pyrogue/protocols/_uart.py

bengineerd:
Threading makes sense. 
We use this on lcls-mr-sbtb2, so perhaps @leosap can make sure it still works?

PR #521, opened 2019-09-24
Update array storage to a dictionary

slacrherbst:
Array variables are now stored as a dictionary, which is returned when the user accesses the variable via direction lookup, allowing the following calls:

````
for v in dummyTree.AxiVersion.TestArray.values():
    v.setDisp('0')

or

for k,v in dummyTree.AxiVersion.TestArray.values():
    print(f"Item {k} = {v}")
````

nodeMatch will still return the matched nodes as an array. A list generation performance penalty is created when slicing arrays in a configuration file.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -257,6 +257,8 @@ class AxiVersion(pr.Device):
 
         self.BuildStamp.addListener(parseBuildStamp)        
 
+        #self.add(pr.LocalVariable(name = 'TestArray[2]',value=0))
+
         for i in range(4):
             for j in range(4):
                 for k in range(4):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_device.py [hunk 1]
```diff
@@ -257,6 +257,8 @@ def parseBuildStamp(var, val):
 
         self.BuildStamp.addListener(parseBuildStamp)        
 
+        #self.add(pr.LocalVariable(name = 'TestArray[2]',value=0))
```
ruck314: Maybe delete the `#self.add(pr.LocalVariable(name = 'TestArray[2]',value=0))` and `#self.add(pr.LocalVariable(name = 'TestArray[2]',value=0))`?

slacrherbst: That is kind of my testing playground. I had those there to verify the overlap checking works write. One of these days I need to script a proper test for this feature.

PR #528, opened 2019-09-27
Generate clear error for unconnected memory devices

slacrherbst:
This PR will ensure an error like below is generated if the user fails to properly connect the memory bus:

pyrogue._Block.MemoryError: 'Memory Error for dummyTree.AxiVersion.FpgaVersion at address 0x000000 Invalid min/max memory interface size. Device or Variable may be unconnected!'

### DIFF
--- a/src/rogue/interfaces/memory/Master.cpp
+++ b/src/rogue/interfaces/memory/Master.cpp
@@ -68,7 +68,7 @@ void rim::Master::setup_python() {
 //! Create object
 rim::Master::Master() {
    error_   = "";
-   slave_   = rim::Slave::create(4,4); // Empty placeholder
+   slave_   = rim::Slave::create(4,0); // Empty placeholder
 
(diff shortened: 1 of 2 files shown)


src/rogue/interfaces/memory/Master.cpp
```diff
@@ -68,7 +68,7 @@ void rim::Master::setup_python() {
 //! Create object
 rim::Master::Master() {
    error_   = "";
-   slave_   = rim::Slave::create(4,4); // Empty placeholder
+   slave_   = rim::Slave::create(4,0); // Empty placeholder
```
ruck314: Why this change?

slacrherbst: It sets the default Slave placeholder module for a master to have an invalid max size which triggers the error in the Block. Otherwise their is no way to tell if the memory device is not connected.

PR #535, opened 2019-10-03
Allow config files to load from zip files

slacrherbst:
This PR allows zip based files and directories to be passed as entries in the LoadConfig filename or filename list. The access occurs just as if the zip file based directory was file system based.

Because of zip file filtering, only files ending in .yml or .yaml will be supported for config loads.

All three of the below commands will load the config.yml file

````
root.LoadConfig('/path/to/zipfile.zip/directory1/')
root.LoadConfig('/path/to/zipfile.zip/directory1')
root.LoadConfig('/path/to/zipfile.zip/directory1/config.yml')
````

### DIFF
--- a/python/pyrogue/_Root.py
+++ b/python/pyrogue/_Root.py
@@ -609,6 +609,7 @@ class Root(rogue.interfaces.stream.Master,pr.Device):
 
         return True
 
+
     def loadYaml(self,name,writeEach,modes,incGroups,excGroups):
         """Load YAML configuration from a file. Called from command"""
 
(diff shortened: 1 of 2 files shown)


ruck314:
Is it possible from the GUI to tunnel down a .zip file to find the dir or path to load?  Or is this only intended for scripting mode?

slacrherbst:
No. Not at this time. I had the goal of doing the actual file browse on the server side which would enable this, as well as solve the client/server file system mismatch, but the gui browse windows make it hard/impossible to implement.

PR #542, opened 2019-10-11
Move Root start params to init method

slacrherbst:
This addresses a nagging problem with how we have been using start() combined with the root context managers. It also addresses the problem where duplicate parameters are passed through the root's init() method to be passed to start.

The parameters passed to the start() method are now passed to Root at creation. Legacy code using start() parameters will still be supported, but with a deprecation warning. 

Also we should no longer call start() in a Root's init method as it is inconsistent with the context manager. Instead the method should be defined in the sub-class to call whatever startup calls need to be made. The context manager __enter__ method will start() the root if not already started and will generate a warning if it detect's the root has been started in init.

Previous root classes:

`````python
class DummyTree(pyrogue.Root):
    def __init__(self):
        pyrogue.Root.__init__(self,name='dummyTree',description="Dummy example")
        self.start(timeout=2.0, pollEn=True, serverPort=0)

        self.epics=pyrogue.protocols.epics.EpicsCaServer(base="test", root=self)
        self.epics.start()

    def stop(self):
        self.epics.stop()
        pyrogue.Root.stop()
`````
Proper method:
`````python
class DummyTree(pyrogue.Root):
    def __init__(self):
        pyrogue.Root.__init__(self,
                                  name='dummyTree',
                                  description="Dummy example",
                                  timeout=2.0,
                                  pollEn=True,
                                  serverPort=0)

        self.epics=pyrogue.protocols.epics.EpicsCaServer(base="test", root=self)

    def start(self):
        pyrogue.Root.start()
        self.epics.start()

    def stop(self):
        self.epics.stop()
        pyrogue.Root.stop()
`````

### DIFF
--- a/python/pyrogue/_Root.py
+++ b/python/pyrogue/_Root.py
@@ -97,16 +97,52 @@ class Root(rogue.interfaces.stream.Master,pr.Device):
 
     def __enter__(self):
         """Root enter."""
+
+        if self._running:
+            print("")
+            print("=======================================================================")
+            print("Detected 'with Root() as root' call after start() being called in")
+            print("Root's __init__ method. It is no longer recommended to call")
+            print("start() in a Root class's init() method! Instead start() should")
+            print("be re-implemented to startup sub-modules such as epics!")
+            print("=======================================================================")
+        else:
+            self.start()
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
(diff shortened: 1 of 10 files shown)


slacrherbst:
I fixed a typo in the description on how the args are passed to Root.__init__

PR #555, opened 2019-10-31
Iterator Update

slacrherbst:
Cleanup of the iterator code and attempt to speedup iterator.

### DIFF
--- a/include/rogue/interfaces/stream/FrameIterator.h
+++ b/include/rogue/interfaces/stream/FrameIterator.h
@@ -64,8 +64,11 @@ namespace rogue {
                // Creator
                FrameIterator(std::shared_ptr<rogue::interfaces::stream::Frame> frame, bool write, bool end);
 
-               // adjust position
-               inline void adjust(int32_t diff);
+               // increment position
+               inline void increment(int32_t diff);
+
+               // decrement position
+               inline void decrement(int32_t diff);
 
             public:
 
--- a/include/rogue/interfaces/stream/FrameIterator.h
+++ b/include/rogue/interfaces/stream/FrameIterator.h
@@ -44,28 +44,25 @@ namespace rogue {
                std::shared_ptr<rogue::interfaces::stream::Frame> frame_;
 
                // Frame position
-               uint32_t framePos_;
+               int32_t framePos_;
 
                // Frame size
-               uint32_t frameSize_;
+               int32_t frameSize_;
 
                // current buffer
                std::vector<std::shared_ptr<rogue::interfaces::stream::Buffer> >::iterator buff_;
 
                // Buffer position
-               uint32_t buffPos_;
+               int32_t buffBeg_;
 
                // Buffer size
-               uint32_t buffSize_;
+               int32_t buffEnd_;
 
                // Current buffer iterator
                uint8_t * data_;
 
-               // Creator
-               FrameIterator(std::shared_ptr<rogue::interfaces::stream::Frame> frame, bool write, bool end);
-
                // adjust position
-               void adjust(int32_t diff);
+               inline void adjust(int32_t diff);
 
             public:
 
--- a/include/rogue/interfaces/stream/FrameIterator.h
+++ b/include/rogue/interfaces/stream/FrameIterator.h
@@ -19,6 +19,7 @@
 **/
 #ifndef __ROGUE_INTERFACES_STREAM_FRAME_ITERATOR_H__
 #define __ROGUE_INTERFACES_STREAM_FRAME_ITERATOR_H__
+
 #include <stdint.h>
 #include <vector>
 #include <cstring>
--- a/include/rogue/interfaces/stream/FrameIterator.h
+++ b/include/rogue/interfaces/stream/FrameIterator.h
@@ -61,11 +61,11 @@ namespace rogue {
                // Current buffer iterator
                uint8_t * data_;
 
-               // adjust position
-               inline void adjust(int32_t diff);
-
                // Creator
                FrameIterator(std::shared_ptr<rogue::interfaces::stream::Frame> frame, bool write, bool end);
+
+               // adjust position
+               inline void adjust(int32_t diff);
 
             public:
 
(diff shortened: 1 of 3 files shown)


russell-slac:
Ryan,

Okay, if you're critic, you have to say something. And that’s about all this is.  None of this is terribly important, except maybe the last point on efficiency.

  *   Why is the construction of the Frame::Iterator done in Frame.cpp?
     *   When I went through FrameIterator.cpp, I could not find where frame_ was initialized, then found it in Frame.cpp
     *    It appears the constructor in FrameIterator.cpp sets some defaults,  with many of its values being overwritten in Frame::beginRead/Write, endRead/Write.
     *   This is just a taste/style argument, so take it or leave it
        *   Why not make a constructor in FrameIterator to be FrameIterator (frame, readWriteFlag)?
     *   I’m sure it is somewhere, could not find the constructor FrameIterator (FrameIterator &)

  *   Again, a style issue that has some efficiency implications
     *   I would have had different adjusts for advancing or retarding the index
     *   Removes half the checks, either
        *   it diff > 0, and one only checks to go to the next buffer
        *   if diff < 0, and one only checks to go to the previous buffer
     *   See below

  *   Even I am not big on this, its just nipping around the edges of increasing efficiency
     *   Only addresses when the boundary is crossed, which, if the frame is reasonably sized, is rare, except in random accesses.

  *   A lot of the complication comes from supporting random access iterators
     *   Would it pay to make specialized iterators for just input and output iterators?

Efficiency Issues
The real efficiency problem is the lowest common denominator approach with bytes.   What if one always guaranteed that a frame had an integral number of 32 or 64-bit values.The iterators would then be templated to be uint8_t, uint16_t, uint32_t or uint64_t.   I just guessing that most accesses are not bytes.  All the above access is the protected by the interface you have defined.  I would consider giving up completely filling a frame to ensure an integral number of 32 or 64-bit values.  Even if you want to completely fill a frame, the reading size can be fixed.  The write side is harder, but doable.


I would also look at how to inline more of this.   There are a lot of calls involved even in internal routines, but again mostly when one crosses frame boundaries.


Now for some pontificating.
This whole C++ iterator model kills almost any optimization that would happen with a good old for loop.  The compiler has no way of using SIMD instructions.  Just can’t recognize the pattern.

Every random access involve a full construction of an iterator via a copy.


JJ


This would look something like:
(The downside is that framePos_ must previously valid.   Of course, if wasn’t, probably makes no sense.  (framePos_ is entirely controlled by your code, so it is not like the user can screw-it

void ris::FrameIterator::adjustPositive(int32_t diff) {
   framePos_ += diff;
   data_     += diff;

   // Underflow or overflow
   if ( framePos_ < 0 || framePos_ >= frameSize_ ) {
      framePos_ = frameSize_;
      buffBeg_  = frameSize_;
      buffEnd_  = frameSize_;
      buff_     = frame_->endBuffer();
      data_     = NULL;
   }

   // We have shifted to a new buffer
   else {

      // Increment current buffer until we find the location of the data position
      // Iterator always contains one extra buffer index
      while ( framePos_ >= buffEnd_ ) {
         buff_++;
         buffBeg_  = buffEnd_;
         buffEnd_ += (write_) ? (*buff_)->getSize() : (*buff_)->getPayload();
      }

      // Set pointer
      data_ = (*buff_)->begin() + (framePos_ - buffBeg_);
   }
}


void ris::FrameIterator::adjustNegagtive(int32_t diff) {
   framePos_ += diff;
   data_     += diff;

   // Underflow or overflow
   if ( framePos_ < 0 || framePos_ >= frameSize_ ) {
      framePos_ = frameSize_;
      buffBeg_  = frameSize_;
      buffEnd_  = frameSize_;
      buff_     = frame_->endBuffer();
      data_     = NULL;
   }

   // We have shifted to a new buffer
   else {


      // Decrement current buffer until the desired frame position is greater than
      // the bottom of the buffer
      while ( framePos_ < buffBeg_ ) {
         buff_--;
         buffEnd_  = buffBeg_;
         buffBeg_ -= (write_) ? (*buff_)->getSize() : (*buff_)->getPayload();
      }
}

      // Set pointer
      data_ = (*buff_)->begin() + (framePos_ - buffBeg_);
   }
}

On Oct 31, 2019, at 12:21 AM, Ryan Herbst <notifications@github.com<mailto:notifications@github.com>> wrote:

@slacrherbst requested your review on: #555 Iterator Update.

slacrherbst:
I need to step through this piece by piece.... but I am certainly open to criticism, its the only way we learn.

* Why is the construction of the Frame::Iterator done in Frame.cpp? 

I had originally done exactly had you proposed with a "friend" constructor called from frame:

`````
FrameIterator(frmae, writeFlag, endflag)
`````

Here I had an if on the write flag. I had thought maybe that would be less efficient, but maybe the right thing to do is create different creators? You are right though, my gain in avoiding the if may be negated by the init and then overwrite.

* I would have had different adjusts for advancing or retarding the index 
* Removes half the checks, either 
* it diff > 0, and one only checks to go to the next buffer 
* if diff < 0, and one only checks to go to the previous buffer * 

This code I have re-written half a dozen times and I am definitely interested in reading through your code below. This is the primary area I am looking for feedback on. With these iterators each instruction is a huge penalty in the loops. 

* A lot of the complication comes from supporting random access iterators 
* Would it pay to make specialized iterators for just input and output iterators? 

This may be the key. Its possible my headaches are from being to generic. Maybe I need to provide different access iterators for different types of operations. 

I will read the rest of your comment tonight/tomorrow. Off to Halloween fun.....

slacrherbst:
JJ,

I moved the init back into the FrameIterator per your suggestion. I also changed the adjust() method a bit to match closer to your suggestion. I kept it in one function because it is hard to determine which direction the adjust() is occurring. Technically someone could write a line of code like:

`````
iter += -5
`````
So I can't easily anticipate if the shift is positive or negative.

slacrherbst:
Per JJ's suggestion I separated increment and decrement functions and ordered the checks for better performance.

PR #556, opened 2019-10-31
Create frame data accessor

slacrherbst:
This creates a templated frame accessor class for random access of frame data in a single buffer range. 

The data can be accessed either with a direct de-reference [] without range checking or with the at() method with range checking.

`````
   ris::FrameAccessor<uint16_t> srcArray(srcIter,copyCount);
   ris::FrameAccessor<uint16_t> dstArray(dstIter,copyCount);

   for (x=0; x < copyCount; x++) dstArray.at(reorgMap[x]) = srcArray.at(x);
`````
or
`````
   for (x=0; x < copyCount; x++) dstArray[reorgMap[x]] = srcArray[x];
`````

### DIFF
--- a/include/rogue/interfaces/stream/FrameAccessor.h
+++ b/include/rogue/interfaces/stream/FrameAccessor.h
@@ -1,0 +1,75 @@
+/**
+ *-----------------------------------------------------------------------------
+ * Title      : Stream frame accessor
+ * ----------------------------------------------------------------------------
+ * File       : FrameAccessor.h
+ * Created    : 2018-03-06
+ * ----------------------------------------------------------------------------
+ * Description:
+ * Stream frame iterator
+ * ----------------------------------------------------------------------------
+ * This file is part of the rogue software platform. It is subject to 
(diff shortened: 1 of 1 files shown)


russell-slac:
Ryan,

Now that I re-found the code

FrameAccessor.h

  *   the increment and decrement methods should be marked private
     *   They are inlined but no implementation is provided, i.e. the user can’t and shouldn’t invoke them


Thinking about whether a interface along the lines of std::transform is useful.

With this you can capture  the details of efficiently iterating within Rogue, while providing the user the ability to control the handling of the  details.

This works for 1 for 1 transformations, but could be problematic t for 1 -> many or many <- 1.   An example of 1 <- many would be something like a convolution where one output is the sum of say 5 inputs.

JJ





On Nov 5, 2019, at 10:53 AM, Ryan Herbst <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #556<https://github.com/slaclab/rogue/pull/556> into pre-release.

PR #561, opened 2019-11-07
Allow clean termination on signal reception

slacrherbst:
This adds signal handlers to the runGui (legacy) and waitCntrl-c helper functions to allow Rogue to terminate gracefully when receiving a SIGTERM signal (kill without the -9).

This will also allow cntrl-c to close the gui properly.

The pydm version (runPyDM) already supports this natively.

### DIFF
--- a/python/pyrogue/__init__.py
+++ b/python/pyrogue/__init__.py
@@ -9,6 +9,7 @@
 #-----------------------------------------------------------------------------
 import sys
 import os
+import signal
 
 MIN_PYTHON = (3,6)
 if sys.version_info < MIN_PYTHON:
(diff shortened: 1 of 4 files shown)


python/pyrogue/__init__.py
```diff
@@ -66,11 +67,23 @@ def addLibraryPath(path):
...
+
+    mon = monitorSignal()
+    signal.signal(signal.SIGTERM, mon.receiveSignal)
+
+    print(f"Running. Hit cntrl-c or send SIGTERM to {os.getpid()} to exit.")
     try:
-        while True:
-            time.sleep(1)
+        while mon.runEnable:
+            time.sleep(0.5)
```
ruck314: why did the sleep change from 1 second to 500 ms?

ruck314:
Is it possible for automatic regression testbed in travis to check for "cntrl-c" or SIGTERM (help with code coverage)?

python/pyrogue/__init__.py
slacrherbst: I wanted the response to be a little more immediate to avoid double kills.

slacrherbst:
Regression testing may be difficult in this case in travis.

PR #564, opened 2019-11-07
Update base FileReader to create a generator.

slacrherbst:
This will break existing uses, but I don't think anyone is using this yet.]

Example usage:

```python
        with FileReader(fileList,configChan=1) as fr:

            for header,data in fr.records():

                 # Process data as uint8 numpy array

                 val = fr.configValue('root.blah.blah')  # Get current config state

```

### DIFF
--- a/python/pyrogue/utilities/fileio/_FileReader.py
+++ b/python/pyrogue/utilities/fileio/_FileReader.py
@@ -45,6 +45,7 @@ class FileReader(object):
         self._header     = None
         self._data       = None
         self._config     = {}
+        self._currFName  = ""
         self._currCount  = 0
         self._totCount   = 0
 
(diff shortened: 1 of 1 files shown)


python/pyrogue/utilities/fileio/_FileReader.py
```diff
@@ -16,74 +16,116 @@
...
-class FileReader(object):
+RogueHeaderSize  = 8
+RogueHeaderPack  = 'IHBB'
 
-    def __init__(self, filename, configChan=None):
-        self._filename   = filename
-        self._configChan = configChan
+# Default header as a named tuple
+RogueHeader = namedtuple( 'RogueHeader',
+                         [ 'size'                ,  # 4 Bytes, uint32, I
```
ruck314: Does size of 32-bit (4 bytes) means there a 4GB limit for the FileReader/FileWriter?  How does the FileReader/FileWriter handle the > 4GB case?

slacrherbst: The 32-bit size field is each record. I hope we don't have records that large.

ruck314: Does the FileWriter close the file and stop when the 4GB limit is reached?

slacrherbst:
The FileWriter can break up the files. But remember the size is each single Frame object written to the file. Do we expect to ever have 4G frames? This is independent of the total file size.

PR #566, opened 2019-11-13
Avoid ZMQ close hangs

slacrherbst:
ZMQ refuses to close properly. This PR avoids hanging on close.


jesusvasquez333:
I tested this fix and solved my problem for the root's `close()` calls hanging

PR #569, opened 2019-11-13
Support nested yaml files

slacrherbst:
This PR adds support for the !include tag in yaml files. The included file will be applied at the dictionary level at which it appears. The combined yaml file follows the same rules as a single yaml file when it comes to repeated tags etc.

Example: test.yml

`````
dummyTree:
  PollEn: True
  AxiVersion: !include AxiVersion.yml
  StreamWriter:
    enable: True
    DataFile: 'test2'
`````
With the included file AxiVersion.yml containing:

`````
ScratchPad: 0x000010
TestArray[6]: 8
TestBool: True
`````
The included file can be relative or absolute, and may be part of a zip file.

### DIFF
--- a/python/pyrogue/_Root.py
+++ b/python/pyrogue/_Root.py
@@ -877,8 +877,17 @@ def yamlToData(stream='',fName=None):
         pass
 
     def include_mapping(loader, node):
-        if fName is not None:
-            filename = os.path.join(os.path.dirname(fName), loader.construct_scalar(node))
+        rel = loader.construct_scalar(node)
+
+        # Filename starts with absolute path
+        if rel[0] == '/':
+            filename = rel
+
+        # Filename is relative and we know the base path
+        elif fName is not None:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ruck314** [comment]
Your example if the `AxiVersion.yml` existed in another directory, could you point to it relative to the base YAML directory like this?
```
dummyTree:
  PollEn: True
  AxiVersion: !include ../anotherDir/AxiVersion.yml
  StreamWriter:
    enable: True
    DataFile: 'test2'
```

slacrherbst:
> Given your example: if the `AxiVersion.yml` existed in another directory, could you point to it relative to the base YAML directory like this?
> 
> ```
> dummyTree:
>   PollEn: True
>   AxiVersion: !include ../anotherDir/AxiVersion.yml
>   StreamWriter:
>     enable: True
>     DataFile: 'test2'
> ```

Yes.

bengineerd (approved):

ruck314 (approved):

PR #571, opened 2019-11-14
Move yaml helping functions to __init__.py

slacrherbst:
This moves some yaml helper functions out of _Root.py into __init__.py since they are more general purpose and not specific to the _Root class.


bengineerd:
Perhaps we should have a `utiliities` module?

slacrherbst:
We already have a utilities package though. Maybe we create a _functions.py?

slacrherbst:
I moved the various functions from __init__.py to _HelperFunctions.py

PR #573, opened 2019-11-14
Deprecate python streamTap and Master::setSlave

slacrherbst:
This PR eliminates the Master::setSlave command, using only addSlave to connect Slaves to a master. The first slave added is considered the "primary" slave and is used for reqFrame calls. Frames are sent to the Slaves in the reverse order of how they were connected.

pyrogue.streamTap will be deprecated and users should just use pyrogue.streamConnect to add additional slaves.

### DIFF
--- a/python/pyrogue/__init__.py
+++ b/python/pyrogue/__init__.py
@@ -108,33 +108,15 @@ def streamConnect(source, dest):
     else:
         slave = dest._getStreamSlave()
 
-    master._setSlave(slave)
+    master._addSlave(slave)
 
 
 def streamTap(source, tap):
     """
-    Attach the passed dest object to the source for a streams
-    as a secondary destination.
-    Connect source and destination stream devices.
-    source is either a stream master sub class or implements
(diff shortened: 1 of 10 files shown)


python/pyrogue/__init__.py
```diff
@@ -108,33 +108,15 @@ def streamConnect(source, dest):
...
-        master = source._getStreamMaster()
-
-    # Is object a native slave or wrapped?
-    if isinstance(tap,rogue.interfaces.stream.Slave):
-        slave = tap
-    else:
-        slave = tap._getStreamSlave()
-
-    master._addSlave(slave)
+    #print("******** Stream TAP is deprecated. Use streamConnect instead. ********")
```
ruck314: This print statement is commented out.  I think you want to uncomment this print() to info the users that streamTap() is deprecated

slacrherbst: I think I want to wait a release or two.

ruck314 (approved):

PR #574, opened 2019-11-14
Deprecate beginWrite and beginRead FrameIterator creators

slacrherbst:
This PR simplifies the FrameIterator use modes by adding Frame::begin() and Frame::end() iterator access methods, deprecating Frame::beginRead() and Frame::beginWrite. In the new use model the user will first specify the payload size of a frame using Frame::setPayload() and then create a single type of iterator, instead of read and write iterators as previously described.

### DIFF
--- a/docs/src/custom_module/sourcefile.rst
+++ b/docs/src/custom_module/sourcefile.rst
@@ -140,12 +140,12 @@ This module is compiled with the :ref:`custom_makefile` described in this sectio
             // Here we request a frame capable of holding 100 bytes
             frame = reqFrame(frameSize_,true);
 
+            // Unlike the python API we must now specify the new payload size
+            frame->setPayload(frameSize_);
+
             // Set an incrementing value to the first 10 locations
             x = 0;
-            for ( it=frame->beginWrite(); it < frame->endWrite(); ++it ) *it = x++;
-
-            // Unlink the python API we must now specify the new payload size
-            frame->setPayload(frameSize_);
+            for ( it=frame->begin(); it < frame->end(); ++it ) *it = x++;
 
             //Send frame
             sendFrame(frame);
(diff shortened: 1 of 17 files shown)


bengineerd:
The naming of `setPayload()` still bugs me. Can we add a `setPayloadSize()` method, and deprecate `setPayload()` so that it still works for a while? Maybe it would be better in a separate PR but I think we should do it.

jesusvasquez333:
I would suggest instead, using a method name following what STL containers do, and rename `setPayload` to `reserve`: http://www.cplusplus.com/reference/vector/vector/reserve/

jesusvasquez333 (approved):
I didn't have a chance to test the new code, but I agree with the general idea of using only `begin` and `end` methods

slacrherbst:
The only problem with using reserve, is it make no distinction between the ray capacity of the frame/buffer combination versus what is considered valid data. reserve would be more consistent with the raw data size, versus the actual size of the contents.

jesusvasquez333:
Yes, correct. But isn't that the case for `setPayload` as well? 

I think all these functinalities are supported already on STL, by a combination of `reserve`, `size` and `capacity` methods.

slacrherbst:
I guess the difference is that setPayload is a way to indicate what portion of the "reserved" data is valid. Where the reserved space can be equal to or greater than the actual data being contained. We would need to find the equivalent call for a std::container, which I am not really seeing.

jesusvasquez333:
Yes, you are right. `reserve` just modify the allocate space, but the size reported by `size` in not affected. The container size changes when new elements are inserted/removed. 

So, `setPayload` would be a sort of `setSize` which doesn't exist (AFAIK) in them.

bengineerd:
But isn't the buffer vs "reserved" size distinction an implementation detail that should be hidden from the user at the API level?

More generally, to what extent does an end user really need to know about the details of Rogue's underlying buffer implementation? Personally, I think it's preferable for the API to basically pretend that 1 Frame = 1 buffer. You could still provide some low level methods to query the gritty details for debugging and profiling.

slacrherbst:
In most cases the user will use the higher level API calls in the Frame and not care about buffers. There are some cases where because of performance you will want to handle the frames at the buffer level to make use of known contiguous memory.

PR #575, opened 2019-11-15
Create common virtual EnableSharedFromThis class

slacrherbst:
This solves a known problem when a class is formed from two classes which utilized the default enable_shared_from_this class. The resulting sub-class will experience bad weak ptr errors. 

Sub-classes formed in c++ would experience a compile error. Sub-classes formed in python would experience a runtime weak ptr error.

### DIFF
--- a/include/rogue/EnableSharedFromThis.h
+++ b/include/rogue/EnableSharedFromThis.h
@@ -1,12 +1,14 @@
 /**
  *-----------------------------------------------------------------------------
- * Title      : Stream memory pool
+ * Title      : Common virtual class to enable shared_from_this() calls.
  * ----------------------------------------------------------------------------
- * File       : Pool.h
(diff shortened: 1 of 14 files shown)


include/rogue/EnableSharedFromThis.h
```diff
@@ -0,0 +1,41 @@
+/**
+ *-----------------------------------------------------------------------------
+ * Title      : Stream memory pool
```
ruck314: This code header looks correct.  This is not the `Pool.h`

slacrherbst: Yes good catch. I meant to go back and fix that after I verified it solved my problem.

slacrherbst: Fixed now.

PR #576, opened 2019-11-15
Allow operators to create stream & memory connections in python and C++

slacrherbst:
This PR enables the use of operators (>>, << & ==) for connecting stream and memory masters and slaves. Each of these operators can be used to replace streamConnect, streamConnectBiDir and busConnect functions. Equivalent operators are also supported in C++.

Example streamConnect replacements:
````python
axiStream >> fifo >> compress >> fileWriter.getChannel(0)

# The reverse direction is also supported
fileWriter.getChannel(0) << compress << fifo << axiStream
````

Example streamConnectBiDir replacement:
````python
axiStream == srp
````

Example busConnect replacement:
````python
memoryMaster >> memorySlave

# The reverse direction is also supported
memorySlave << memoryMaster
````

These connection methods also support python classes which serve as containers for stream and memory devices, as long as accessor and operator override methods are provided. As an example the following methods are included with the PrbsPair class to enable use of the >>, << and == stream operators.

````python
    def _getStreamMaster(self):
        return self.PrbsRx._getStreamMaster()

    def _getStreamSlave(self):
        return self.PrbsTx._getStreamSlave()

    def __rshift__(self,other):
        pyrogue.streamConnect(self,other)
        return other

    def __lshift__(self,other):
        pyrogue.streamConnect(other,self)
        return other

    def __eq__(self,other):
        pyrogue.streamConnectBiDir(other,self)
````

The streamConnect and streamConnectBiDir are included as helper functions. These functions are required to support both native stream and memory classes as well as generic classes which contain stream and memory classes, using the _getStreamMaster, _getStreamSlave, _getMemoryMaster and _getMemorySlave to gain access to the contained objects.

When using the C++ versions of the operators, you need to de-reference the left hand side:

````c
*mast >> slave;

// or for chained connections
*(*axiStream >> fifo) >> slave;
````

### CODE REVIEW

File: docs/src/interfaces/memory/connecting.rst [hunk 1]
```diff
@@ -11,33 +11,22 @@ A memory master and slave are connected using the following commands in python:
    import pyrogue
 
    # Assume we are crated a memory tree with two masters, masterA and masterB
```
ruck314: Change `crated ` to `creating`

docs/src/interfaces/memory/connecting.rst
```diff
@@ -11,33 +11,22 @@ A memory master and slave are connected using the following commands in python:
...
+   # Connections can also go in the reverse order
+   srpV3 << hubB << masterB
 
 The equivelent code in C++ is show below:
 
 .. code-block:: c
 
-   #include <rogue/Helpers.h>
-
    // Assume we are crated a memory tree with two masters, masterA and masterB
```
ruck314: Change `crated ` to `creating`

docs/src/interfaces/stream/connecting.rst
```diff
@@ -38,15 +36,14 @@ it is passed to the primary slave:
...
 And in C++:
 
 .. code-block:: c
 
-   #include <rogue/Helpers.h>
-
-   rogueStreamConnect(myMaster, mySlave)
+   *myMaster >> mySlave;
 
 In some cases rogue entties can serve as both a stream Master and Slave. This is often the case when
```
ruck314: Change `entties` to `entities`

docs/src/interfaces/stream/debugStreams.rst
```diff
@@ -6,7 +6,7 @@ Debugging Streams
 
 The base stream :ref:`interfaces_stream_Slave` class can server as a debug Frame receiver. This can
 be usefull when you want to display some or all of the data streams used in your designs. In most
-cases the debug Slave module will be used as a stream tap. In order to enable debugging the setDebug()
+cases the debug Slave module will be used as a secndary slave. In order to enable debugging the setDebug()
```
ruck314: Change `secndary` to `secondary`

docs/src/interfaces/stream/debugStreams.rst
```diff
@@ -6,7 +6,7 @@ Debugging Streams
 
 The base stream :ref:`interfaces_stream_Slave` class can server as a debug Frame receiver. This can
 be usefull when you want to display some or all of the data streams used in your designs. In most
```
ruck314: Change `usefull` to `useful`

docs/src/interfaces/stream/usingFifo.rst
```diff
@@ -122,11 +111,8 @@ Below is the equivalent code in C++
...
    # Connect the src and dst
-   rogueStreamConnect(src, dst);
-
-   # Add the Fifo as a stream tap
-   rogueStreamTap(src, fifo);
+   src >> dst;
 
-   # Connect the monitor to the FIfo output
-   rogueStreamConnect(fifo, mon)
+   # Add the Fifo as a second stream and on to the monitor (reverse order show as an example)
```
ruck314: maybe add a comment that the FIFO will assert back pressure on the mon or maybe change this example to a FIFO that will not back pressure on a data path (a lot of time you don't want monitors (like live displays) to back pressure the stream source

include/rogue/interfaces/memory/Slave.h
```diff
@@ -176,6 +177,18 @@ namespace rogue {
                 * @param transaction Transaction pointer as TransactionPtr
```
ruck314: Change `transaction Transaction` to `Transaction` (delete repeated word)

include/rogue/interfaces/stream/Master.h
```diff
@@ -39,7 +44,7 @@ namespace rogue {
           * interfaces to one or more stream slave objects. The frst stream Slave is used
```
ruck314: Change `frst` to `first`

include/rogue/interfaces/stream/Master.h
```diff
@@ -39,7 +44,7 @@ namespace rogue {
           * interfaces to one or more stream slave objects. The frst stream Slave is used 
           * to allocated new Frame objects and it the last Slave to receive frame data.
```
ruck314: Change `objects and it the last Slave` to `objects, and it is the last Slave`

tests/test_memory.py
```diff
@@ -72,7 +73,8 @@ def __init__(self):
 
         # Create a memory gateway
         self.ms = rogue.interfaces.memory.TcpServer("127.0.0.1",9080);
-        pr.busConnect(self.ms,self.sim)
+        #self.ms >> self.sim
```
ruck314: Do you want to remove the commented code?   Or add another part in the test code that test `>>` (other direction) of the bus connecting?

ruck314 (approved):

bengineerd:
Is there no way around having to dereference the left side when working in C++?
```c
*master >> slave;
```
It feels like the kind of thing that will constantly trip people up.

Is it still the left side when assigning in the other direction?
```c
*slave << master;
```

slacrherbst:
In order to avoid that we would have to subclass the smart pointers, and some of the oeprators such as == would generate additional problems. It is always the left side since it is the entity which is being "called".

PR #587, opened 2019-12-07
Pcas from cmake arg

bhill-slac:
Get EPICS_PCAS_ROOT from cmake arg instead of env variable.

The PCAS libs aren't needed for any users environment so no users will
have an EPICS_PCAS_ROOT env variable.   Better to pass as a cmake arg
so during rogue development we can easily specify the pcas version.

Also show separate msg for Found PCAS vs Found EPICS and change
the EPICSV3_* cmake variables to EPICS_*.


slacrherbst:
This breaks the current travis build and will also break the anaconda builds. Perhaps a better approach would be to allow the user to override the environment variable search if the path is passed on the command line.

jesusvasquez333:
Yes, I think being able to use the current environmental variable but at the same time being able to override it with a cmake argument is a good compromise.

ruck314:
@bhill-slac: here's where your PR build during Travis CI:
https://travis-ci.org/slaclab/rogue/builds/624021273?utm_source=github_status&utm_medium=notification

slacrherbst:
Superseded by #593

PR #588, opened 2019-12-12
Fix saveAddressMap function.

bhill-slac:
Function def wrong.  Changed root to self.


slacrherbst:
This has been addressed already in pre-release. I will do a release today.

PR #589, opened 2019-12-12
PCAS from cmake arg

bhill-slac:
Set EPICS_PCAS_ROOT via cmake arg instead of env variable.
Also changed all EPICSV3_ to EPICS_ as V3 is no longer ubiquitous.


ruck314:
PR for this already existing (PR https://github.com/slaclab/rogue/pull/587). Closing ticket

PR #599, opened 2019-12-20
Make base path determination more robust in _HelperFunctions.py

bhill-slac:
When embedding pyrogue, sys.argv is empty.
  File "/reg/g/pcds/epics-dev/bhill/modules/ADPgpCamlink-git/pgpCamlinkApp/pythonSrc/pgpCamLink.py", line 11, in <module>
    import setupLibPaths
  File "/reg/neh/home/bhill/git-wa-neh/slaclab/cameralink-gateway-git/software/scripts/setupLibPaths.py", line 16, in <module>
    pr.addLibraryPath(top_level+'firmware/submodules/axi-pcie-core/python')
  File "/reg/neh/home/bhill/git-wa-neh/slaclab/rogue-git/python/pyrogue/_HelperFunctions.py", line 29, in addLibraryPath
    base = os.path.dirname(sys.argv[0])
IndexError: list index out of range

### DIFF
--- a/python/pyrogue/_HelperFunctions.py
+++ b/python/pyrogue/_HelperFunctions.py
@@ -26,7 +26,7 @@ def addLibraryPath(path):
     Passed strings can either be relative: ../path/to/library
     or absolute: /path/to/library
     """
-    base = os.path.dirname(sys.argv[0])
+    base = os.path.dirname(os.path.realpath(__file__))
 
     # If script was not started with ./
     if base == '': base = '.'


slacrherbst:
The reason I used sys.argv[0] is that __file__ does not work properly when it comes to symbolic links in file systems. Using sys.argv[0] ensures that path is always relative. This is why the next few lines look at for path starting with a . and/or absolute paths. Maybe we should check if sys.argv is empty and use __file__ if it is.

bhill-slac:
Isn't the symbolic link issue resolved by the call to os.path.realpath()?

slacrherbst:
The problem is the inverse.  Where __file__ may return the absolute path when setting up library paths you need everything to be relative to the current location. If I recall directly the problem popped up with the updatePcieFpga script.

slacrherbst:
In general you need to know both the current working directly and also need to know how the program was executed. What is an example of a situations where the argv is not loaded?

slacrherbst:
How about this:
    
    if len(sys.argv) == 0:
        base = os.path.dirname(os.path.realpath(__file__))

    else:
        base = os.path.dirname(sys.argv[0])

        # If script was not started with ./	    # If script was not started with ./
        if base == '': base = '.'

bhill-slac:
The above works for my embedded rogue build.    I'm still not clear on the error you were seeing w/ soft links.   Should be ok unless you're following a path outside your git workspace.

slacrherbst:
Can you edit your PR and I will merge?

slacrherbst:
Addressed in #606

PR #604, opened 2020-01-21
adding placeholders for docs

ruck314:
## Description
- docs/src/index.rst not changed to prevent releasing while still in draft form
- adding placeholders for docs/src/protocols
- adding placeholders for docs/src/utilities


slacrherbst:
In the future doc changes can be directly merged or pushed into the documentation branch.

PR #613, opened 2020-01-25
Structural Changes Creating C++ Block Class, and improving variable update processing

slacrherbst:
The following changes are made to the rogue code in order to support a C++ based Block class and the ability to pre-define the Block size and/or create a custom block class. Other changes are also made in an attempt to reduce the number python class for each set/get operation. The processing time for variable update queuing is also improved.

1. The Block class now performs the bulk of the data conversions previously performed by the toBytes() and fromBytes() calls in the Model class. A blockFunc field in the Model class now defines which Block function is used for the particular model. A Model may set blockFunc=PyFunc to continue to support a python based toBytes and fromBytes call. The Block class may be sub-classed and a user can use the Custom base constant to define custom conversions in the Block class.

2. Range and data validation checking is now performed in the Block class. 

3. The blockSize parameter is removed from the Device class. In the new model the user will manually create a larger block by creating the block directly and adding it to the device using the device.addCustomBlock() method which is used during the buildBlocks call. This has the benefit of no requiring all Device blocks to be the same size. This larger block can be created using the default MemoryBlock class in the current python version. 

4. The C++ Block (and LocalBlock) store a cached version of the Device's enable flag in order to improve per transaction efficiency. The Block's enable state is changed in the Device's enableChanged callback which is called anytime the enable state changes. This optimization greatly improves the measured transaction throughput.

5. The buildBlocks() call in the Device is updated to support the pre-made blocks as well as move some calls out of the Block class. 

6. Some changes are made to the Variable class to better differentiate the set and get methods for the various sub-classes. Some redundant code is generated but removes some calls to if statements. For the RemoteVariable types the conversion to a byte array occurs in the variable class where before it occurred in the Block class. This will be necessary in order to support the C++ block class.

7. The use of the root.updateGroup and root._queueUpdate is updated so that the updateGroup() context is no longer created on every call to checkTransaction. This reduces a large processor overhead. The result is single transactions will still properly queue their updates and the user is still allows to group updates in higher level code using the with root.updateGroup() call. Similarly the distribution of calls between the updateGroup(), _queueUpdate and update thread is changed to increase efficiency. 

8. The grouping of update calls is now also done per thread, allowing updates from one thread to continue to be processed while another thread queue's up its updates. 

9. The ability to pass a variable list to writeBlocks, verifyBlocks, readBlocks and checkBlocks is now removed. Only a single variable can be passed to these functions. This change was made to speed up the set and get calls, sacrificing a seldom used feature for higher performance.


slacrherbst:
Please re-review this PR. I have merged in the latest changes which move the data conversion functions from python to C++ while still allowing python based conversions for corner cases.

slacrherbst:
There are two incomplete portions of Block.cpp that still need to be updated:
1. The bitReverse function needs to be implemented.
2. The FixedPoint set and get functions need to be checked/fixed.

PR #615, opened 2020-01-29
Add ability to build both STATIC_LIB and SHARED_LIB

bhill-slac:
Allows both STATIC_LIB and SHARED_LIB to be defined thus building both variants if desired.
Only for rogue-core.


bhill-slac:
Not tested on macos.

PR #626, opened 2020-02-14
Add checks for payload not set and payload too small in toFrame

bhill-slac:
Output when payload not set:
ClSerialMaster::sendBytes: Frame size 20, payload 0, remBuffer 0, avail 20
terminate called after throwing an instance of 'rogue::GeneralError'
  what():  FrameIterator::toFrame: General Error: Payload size not set
  Aborted (core dumped)

Output when payload size too small:
ClSerialMaster::sendBytes: Frame size 20, payload 5, remBuffer 0, avail 15
terminate called after throwing an instance of 'rogue::GeneralError'
  what():  FrameIterator::toFrame: General Error: Frame too small to accept all data
  Aborted (core dumped)

### DIFF
--- a/include/rogue/interfaces/stream/FrameIterator.h
+++ b/include/rogue/interfaces/stream/FrameIterator.h
@@ -23,6 +23,8 @@
 #include <stdint.h>
 #include <vector>
 #include <cstring>
+#include "rogue/GeneralError.h"
+#include "rogue/interfaces/stream/Frame.h"
 
 namespace rogue {
    namespace interfaces {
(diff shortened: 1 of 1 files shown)


include/rogue/interfaces/stream/FrameIterator.h
```diff
@@ -106,6 +108,15 @@ namespace rogue {
                 */
                uint32_t remBuffer();
 
+               //! Get payload size
```
slacrherbst: I am not a fan of this and only adds a small amount of information to the below error.

include/rogue/interfaces/stream/FrameIterator.h
```diff
@@ -254,6 +265,14 @@ namespace rogue {
                iter += csize;
                size -= csize;
             } while ( size > 0 && csize > 0 );
+#if 1
+			if ( size > 0 ) {
```
slacrherbst: I think we can leave this at:

if ( size > 0 ) throw(rogue::GeneralError("FrameIterator::toFrame","Frame to small to accept all data");

slacrherbst:
We will not add this.

PR #632, opened 2020-02-20
Add Variable Class in C++, performance improvements

slacrherbst:
This PR creates a new C++ class Variable which becomes a parent of Remote Variable. This allows for performance improvements in the python interface as well as creating a C++ API for accessing Rogue variables.

Additional changes to be aware of:

1. Stale state tracking in the Block is now done at both the block and variable level, instead of at the bit level. The previous shadow memory has been removed to avoid unnecessary copies. Instead read and verify are disabled for blocks that are stale. When a variable.set() occurs the variable associated with the set and the parent block are both set stale. The next write or post transaction will ensure the block memory associated with all stale variables is written and the block and the variable state state will be cleared. The stale state of a block should not persist for very long, so the read disable should not be a problem. 

2. The interface to startTransaction now takes an optional variable pointer instead of the previous lowByte and highByte attributes. This avoids a mapped lookup at the C++ layer.

3. set() and get() calls to the block are now done through the variable class directly. 

4. An updateEn flag is now added to variables with a default value of True. Setting this to false will disable the variable tree update calls when a variable is accessed. This is useful when you have variables that are not associated with GUIs or other interfaces.  This will also allow normal variable to take the place of rawWrite and rawRead in most cases. 

5. Functions used by the block to convert python objects to c++ types are now stored as function pointer variables in the variable class. This allows a sub-class to override these.

6. The C++ API to variable set and get calls have function pointers for each interface type allowing block sub-class to define special conversion functions.


bengineerd:
`updateEn` is a bit ambiguous of a name. Perhaps `broadcastUpdates`? Or something like that?

python/pyrogue/_Variable.py
```diff
@@ -633,13 +630,11 @@ def post(self,value):
...
         """
-        self._log.debug("{}.post({})".format(self, value))
-
         try:
 
             # Set value to block
-            self._block.set(self, value)
-            self._block.startTransaction(rogue.interfaces.memory.Post, False, True, 0, -1)
+            self._set(value)
+            self._block.startTransaction(rim.Post, False, True, self)
```
bengineerd: It would be better to include the argument names in these calls with multiple booleans.

slacrherbst: Historically I have not done that with cpp calls, but it appears you may be able to defined named variables in the cpp interface. I will take a look.

bengineerd:
> The stale state of a block should not persist for very long, so the read disable should not be a problem.

That sounds ripe for a corner case.

python/pyrogue/_Variable.py
bengineerd: Crap, I forgot this was in C++.

In the case of `check` maybe it's better just to require a separate explicit call to `checkTransaction()`. It's more readable that way.

What is the use case for forced writes again?

slacrherbst: I was wondering that was well. I was tempered to get rid of it.

slacrherbst: Forced wries occur when the variable ins directly written to. We could detect that when the passed variable is not null

slacrherbst:
> > The stale state of a block should not persist for very long, so the read disable should not be a problem.
> 
> That sounds ripe for a corner case.

I think it should be ok. Reads will just be ignored for a small window. Stale gets cleared even when a device is disabled. I need to look at how default values are handled though.

slacrherbst:
> `updateEn` is a bit ambiguous of a name. Perhaps `broadcastUpdates`? Or something like that?

How about updateNotify? This is tricky but it will disable all updates, like the gui etc.

python/pyrogue/_Variable.py
slacrherbst: Force gets used on global writes, so it does need to be a separate arg.

slacrherbst: I added comments anytime the startTransaction is called.

PR #635, opened 2020-02-25
Add warning when transaction timers are refreshed

slacrherbst:
This will generate a warning every timeout period when a transaction timer is refreshed. This will ensure some indication is present on the console in situations where data flow slows down enough to slow the user interface, but now slow enough to cause timeouts.

This also fixes a bug where the allocation counter was decremented in the wrong place, this is for tracking purposes only and had to real impact on system operation.


slacrherbst:
Please note that in the current and previous implementations the transaction time is not started until the slave accepts the transaction. I think that is ok.

PR #636, opened 2020-02-26
Fix reverse bits processing

slacrherbst:


bengineerd:
It probably makes sense to put this straight into master and cut a patch release right away.

PR #640, opened 2020-02-28
Add base class for creating pure C++ device driver libraries

slacrherbst:
This PR creates a new class, LibraryBase, which can be used as a base class for creating a pure C++ device driver using a Rogue tree.

The steps will be the following:

1. Make sure all Slave memory interfaces are named with the slave.setName() call. 

2. Export Rogue RemoteVariable instances into a header file with the following python command:
    root.genAddressMap('MyHeader.h',headerEn=True)

3. Create a new sub-class of LibaryBase which recreates the stream and memory structures to match the python root. Stream interfaces which will be used by external programs should be declared as public variables in the sub-class. The memory interfaces are defined using the method addMemory("name",slave); The passed names should match the names as present in the exported address map.

4. In the sub-class include the header exported in step 2 and call the following method in the sub-class creator:
    parseMemMap(ROGUE_ADDR_MAP);

5. Compile the code as s shared library.

The user code using this created library can use the methods getVariable("path") and getVariableList() methods to gain access to the Variable() classes in the system. 

This that we may need to tweak as we begin to to use this are:
   1. Does the name LibraryBase make sense?
   2. Do the C++ interfaces to the Variable class make sense?


bhill-slac:
So if I understand your proposal, the C++ API would be extended to add support for calls like this:
  Variable linkReady = getVariable( "ClinkDev.ClinkFeb[0].PgpMon[0].RxRemLinkReady" );
That would be great as would avoid the need to export the address map and parse it.

Would this require any C++ code in projects like cameralink-gateway?
Or would a call like this be added to cameralink-gateway and the resulting header be used by my C++ code?
- root.genAddressMap('MyHeader.h',headerEn=True)

While the base class name doesn't really matter, I think something like rootAddressMap would be more intuitive than LibraryBase.

PR #653, opened 2020-03-11
Allow auto-cleanup of threaded interfaces on Root exit

bengineerd:
So far I've added the following:
 - A `stop()` method to `stream.Master, stream.Slave, memory.Master and memory.Slave`
 - 'stop()' methods to the appropriate protocols
 - A `addInterface()` and 'addProtocol' method to `Device`
 - `Device.stop()` automatically calls stop on every interface that has been added

A few notes:
 - Adding functionality to `Node.add()` is a bit difficult, as it really expects a Node object in a lot of ways. A new method `addInterface` that only applies to `Devices` seemed cleaner from an implementation perspective.
 - I settled on `stop` instead of `close`.
 - I've implemented `stop()` for the UDP protocol stuff. Will eventually add every rogue interface that has an internal thread
 - Also haven't touched the memory interfaces but in theory those might want a `stop()` method as well.
 - I'm pretty sure I did something wrong trying to expose `stop()` to python.


slacrherbst:
I added some further updates to this and added the addProtocol() methods which is identical to addInterface() but is less confusing to the user when adding a protocol.

include/rogue/interfaces/ZmqClient.h
```diff
@@ -65,7 +65,7 @@ namespace rogue {
 
             std::string send(std::string value);
 
-            void close();
+            void stop();
```
ruck314, line 68: Will removing the close() break any existing builds or interfaces?

python/pyrogue/protocols/_uart.py
```diff
@@ -30,7 +30,11 @@ def __init__(self, device, baud, timeout=1, **kwargs):
         self._workerThread = threading.Thread(target=self._worker)
         self._workerThread.start()
 
+    # Deprecated
     def close(self):
+        self.stop()
```
ruck314, line 35: Need to add a print statement about how  `close()` is deprecated and will be removed in the future and should use `stop()` instead

src/rogue/interfaces/memory/TcpClient.cpp
```diff
@@ -110,10 +110,15 @@ rim::TcpClient::TcpClient (std::string addr, uint16_t port) : rim::Slave(4,0xFFF
 
 //! Destructor
 rim::TcpClient::~TcpClient() {
-  this->close();
+  this->stop();
 }
 
+// deprecated
 void rim::TcpClient::close() {
+   this->stop();
```
ruck314, line 118: Need to add a print statement about how  `close()` is deprecated and will be removed in the future and should use `stop()` instead

src/rogue/interfaces/memory/TcpServer.cpp
```diff
@@ -103,10 +103,14 @@ rim::TcpServer::TcpServer (std::string addr, uint16_t port) {
 
 //! Destructor
 rim::TcpServer::~TcpServer() {
-  this->close();
+  this->stop();
 }
 
 void rim::TcpServer::close() {
+    this->stop();
```
ruck314, line 110: Need to add a print statement about how  `close()` is deprecated and will be removed in the future and should use `stop()` instead

src/rogue/interfaces/stream/TcpCore.cpp
```diff
@@ -131,10 +131,15 @@ ris::TcpCore::TcpCore (std::string addr, uint16_t port, bool server) {
 
 //! Destructor
 ris::TcpCore::~TcpCore() {
-  this->close();
+  this->stop();
 }
 
+// deprecated
 void ris::TcpCore::close() {
+   this->stop();
```
ruck314, line 139: Need to add a print statement about how  `close()` is deprecated and will be removed in the future and should use `stop()` instead

include/rogue/interfaces/ZmqClient.h
slacrherbst: No. ZmqClient is only really used with the VirtualClient class and close() was never actually ever called outside the destructor.

python/pyrogue/protocols/_uart.py
slacrherbst: I think this will flush out over time and the calls to close() won't cause any issues.

src/rogue/protocols/packetizer/Controller.cpp
```diff
@@ -62,7 +62,9 @@ rpp::Controller::Controller ( rpp::TransportPtr tran, rpp::ApplicationPtr * app,
 }
 
 //! Destructor
-rpp::Controller::~Controller() { }
+rpp::Controller::~Controller() {
+   this->stopQueue();
```
ruck314, line 66: There is a `stopQueue()` and `stop()` for this module.  Should the `stop()` be used in the destructor instead?  This assumes that `stopQueue()` is executed inside `stop()`

tests/test_memory.py
```diff
@@ -65,16 +65,19 @@ def __init__(self):
                          serverPort=None)
 
         # Use a memory space emulator
-        self.sim = pr.interfaces.simulation.MemEmulate()
+        sim = pr.interfaces.simulation.MemEmulate()
+        self.addInterface(sim)
```
ruck314, line 69: Are future rogue scripts going to have to use `self.addInterface()` to add interfaces to the py.root or will the standard method that we use everywhere still work?

src/rogue/protocols/packetizer/Controller.cpp
slacrherbst: I almost went down that road, but this is an inner class and the cleanup is handled at a higher level. so stopQueue() is just an internal call. It also does not have a stop() call.

tests/test_memory.py
slacrherbst: addInterface() is only really a helper. existing scripts will not break.

bengineerd:
I will integrate this into my KPIX setup today to test it out.

slacrherbst:
We should decided on stop() or _stop() for the function we want to use for this.

PR #658, opened 2020-05-25
Escape \n and " characters when generating dump header.

bhill-slac:
Was able to compile and parse cameralink-gateway header file w/ this fix.

### DIFF
--- a/python/pyrogue/_Root.py
+++ b/python/pyrogue/_Root.py
@@ -583,7 +583,13 @@ class Root(rogue.interfaces.stream.Master,pr.Device):
                 data += "{}\t".format(v._block._reqSlaveName())
                 data += "{}\t".format(v._block.path)
                 data += "{:#x}\t".format(v._block.size)
-                data += "{}".format(v.description)
+                # Escape " characters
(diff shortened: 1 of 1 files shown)


bengineerd:
It would be helpful to know what error you were getting, and what input produced it. It also seems like `str.replace()` would be a bit simpler way to do this. 

This stackoverflow QA seems to cover what you are trying to do. https://stackoverflow.com/questions/18886596/replace-all-quotes-in-a-string-with-escaped-quotes.

bhill-slac:
> 
> 
> Which Device or Variable has `"` and `\n` characters in its description?
AlertMaskRead in surf/python/surf/devices/nxp/_Sa56004x.py was first one w/ \n but there are many in cameralink-gateway project.
FrameMode in surf/python/surf/protocols/clink/_ClinkChannel.py had an unescaped " and
LinkMode in surf/python/surf/protocols/clink/_ClinkChannel.py had two escaped ".

bhill-slac:
str.replace works for escaping ", but we need the other solution to strip internal whitespace in multi-line descriptions.

PR #662, opened 2020-06-01
#ifndef NO_PYTHON build fixes

bhill-slac:
Required to build w/ NO_PYTHON.
Am I the only one building NO_PYTHON???


slacrherbst:
I don't think I compiled with NO_PYTHON after re-writing this code.

PR #660, opened 2020-05-30
Travis CI Updates

ruck314:
### Description
- Resolved the python Linter bugs in the current pre-release branch
- Updating to Ubuntu 20.04 LTS distro
- Use 'apt install libboost-all-dev` instead of building for source

PR #664, opened 2020-06-02
Enable blocks added to Library wrapper

slacrherbst:


python/pyrogue/utilities/fileio/_FileReader.py
```diff
@@ -92,7 +92,7 @@ def _nextRecord(self):
                 try:
                     self._data = numpy.fromfile(self._currFile, dtype=numpy.int8, count=payload)
                 except Exception:
-                    raise rogue.GeneralError(f"fileio.FileReader","Failed to read data from {self._currFname}")
+                    raise rogue.GeneralError("fileio.FileReader","Failed to read data from {self._currFname}")
```
bengineerd, line 95: Doesn't `"Failed to read data from {self._currFname}"` need to be an f-string.

ruck314: https://github.com/slaclab/rogue/pull/660/commits/0ff511a4795626a42074487fe954adbdbda7d936
Fixed in the travis-focal branch

bhill-slac:
Tested with pgp framegrabber and works as expected.

PR #668, opened 2020-06-03
Updating to Ubuntu 20.04 LTS distro and CMAKE Update

ruck314:
### Description
- Ubuntu 20.04 LTS distro
- Use 'apt install libboost-all-dev` instead of building for source
- Updated CMakeLists for using apt install of boost in travis CI
  - Based on this [CMakeLists.txt](https://github.com/TNG/boost-python-examples/blob/master/CMakeLists.txt)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: focal
 sudo: required
 
 branches:
@@ -10,7 +10,10 @@ addons:
     packages:
       - doxygen
       - doxygen-doc
-      - python-dev
+      - python3-dev
+      - python3-pip
+      - libzmq3-dev
+      - libboost-all-dev
 
 services:
   - docker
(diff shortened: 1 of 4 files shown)


slacrherbst:
Ok. so this worked. I am worried that removing the existing cmake chain may break other builds.

ruck314:
The nice thing about this  new CMakeLists.txt method for boost is we no longer have to update it everytime there's a new release of python (e.g. python3.9 would force us to modify the CMakeLists.txt again)

slacrherbst:
Yes I think it is the right approach with the variables but it fails for non ubuntu boost python versioning. I am going to tweak it a bit. Also not that the file templates/RogueConfig.cmake.in should also be updated to match the core CMakelists.txt. It is used when compiling libraries against rogue.

ruck314:
Feel free to tweek it.  I think it's the right direction to go on this

slacrherbst:
See my commit above. This builds off of Larry's changes but adds back the varieties of library naming formats from the previous CMakefile. This also adds a helper for anaconda based builds.

PR #679, opened 2020-06-22
Addresses a corner case where a false timeout can occur with the TcpClient

slacrherbst:
The TcpClient memory class was not adding its transaction to the tracking queue until after the transaction was sent. This resulted in a corner case where the transaction could be lost if it is received too quickly.

This also removes a possible cascading timeout error when cleaning up the transaction tracking map.

This also removes noisy creation debug messages from the transaction class.

### DIFF
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -17,7 +17,7 @@ import pyrogue.interfaces.simulation
 import rogue.interfaces.memory
 import time
 
-#rogue.Logging.setLevel(rogue.Logging.Debug)
+#rogue.Logging.setLevel(rogue.Logging.Warning)
 #import logging
 #logger = logging.getLogger('pyrogue')
 #logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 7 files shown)


tests/test_memory.py
```diff
@@ -69,6 +69,7 @@ def __init__(self):
 
         # Create a memory gateway
         self.ms = rogue.interfaces.memory.TcpServer("127.0.0.1",9080);
+        time.sleep(1)
```
ruck314, line 72: Instead of a `time.sleep()`, can the `rogue.interfaces.memory.TcpServer()` have a variable that you can poll to know when the link is up?  Example of what I would use instead of time.sleep
```
while (self.ms.LinkUp != 0x0)
```

slacrherbst: In reality this should not be necessary.

slacrherbst:
Passes travis with all sleep statements removed.

ruck314 (approved):

PR #680, opened 2020-06-25
Pass exceptions on to callers

slacrherbst:
This PR changes the Rogue approach to handling errors. Previously the errors were caught and logged. In this new model the exceptions are passed on to the caller. This allows the caller to become aware of the errors and avoids some errors being masked, sometimes resulting a sequence of commands continuing event after a fatal error occurs.

With this change, exceptions are now passed across the ZMQ client/server link.

The legacy pyqt GUI will simply log an error. The PYDM interface will open an error dialog box. 

I am still looking at the proper way to deal with exceptions called from the epics3 and epics4 interfaces.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,12 +112,13 @@ jobs:
       script:
         # Run tests with coverage
         - cd $TRAVIS_BUILD_DIR
-        - python3 -m pytest --cov
+        - python3 -m pytest
+        #- python3 -m pytest --cov
 
       after_success:
         # Upload coverage report
-        - codecov
-        - coverage report -m
+        #- codecov
+        #- coverage report -m
         # Publish docs.
         - if [ "$TRAVIS_BRANCH" = "documentation" ]; then
            doctr deploy . --built-docs docs/build/html --deploy-branch-name gh-pages --branch-whitelist documentation;
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ language:
   - python
 
 python:
-  - 3.6
+  - 3.7
 
 stages:
   - test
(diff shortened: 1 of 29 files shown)


.travis.yml
```diff
@@ -112,12 +112,13 @@ jobs:
       script:
         # Run tests with coverage
         - cd $TRAVIS_BUILD_DIR
-        - python3 -m pytest --cov
+        - python3 -m pytest
+        #- python3 -m pytest --cov
```
ruck314, line 116: How come this is committed out?

.travis.yml
```diff
@@ -112,12 +112,13 @@ jobs:
...
 
       after_success:
         # Upload coverage report
-        - codecov
-        - coverage report -m
+        #- codecov
```
ruck314, line 120: How come this is committed out?

.travis.yml
slacrherbst: It appears to be broken in the master branch version of the build system. There is a version conflict which will require upgrading travis, which is already in place for pre-release. I figured just to temp get by for now.

tests/test_server.py
```diff
@@ -170,10 +172,10 @@ def _myArray(self):
     with DummyTree() as dummyTree:
         dummyTree.saveAddressMap('test.csv')
 
-        pyrogue.waitCntrlC()
+        #pyrogue.waitCntrlC()
```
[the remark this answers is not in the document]
slacrherbst: that was a mistake. I corrected it.

jesusvasquez333:
I'm going to test this one in one of the SMuRF system...

jesusvasquez333:
I tested this changes yesterday and I was able to catch exception thrown during the YAML defaults loading.

PR #682, opened 2020-06-26
Fix pydm tree column size adjustment

slacrherbst:


python/pyrogue/pydm/widgets/command_tree.py
```diff
@@ -139,6 +154,12 @@ def __init__(self,*,path,top,parent,command):
 
             self._top._tree.setItemWidget(self,3,self._widget)
 
+            width = fm.width('0xAAAAAAAA    ')
```
ruck314, line 157: `0xAAAAAAAA    ` seems strange.  Do you need all that white-space in the python string argument?

python/pyrogue/pydm/widgets/variable_tree.py
```diff
@@ -182,6 +198,11 @@ def __init__(self,*,path,top,parent,variable):
 
         self._top._tree.setItemWidget(self,3,w)
 
+        width = fm.width('0xAAAAAAAA    ')
```
ruck314, line 201: 0xAAAAAAAA seems strange. Do you need all that white-space in the python string argument?

PR #686, opened 2020-07-02
Fix network wrapper start call in RSSI

slacrherbst:


ruck314:
@slacrherbst Here are all the `.start()` in the pre-release:
image

Other than `_Network.py` (that's fixed in this PR), do any of these other `.start()` need to be changed to `._start()`?

slacrherbst:
The rest of these are either built in methods (python threads) or methods that are meant to be called by the user.

PR #688, opened 2020-07-02
Const refs

bhill-slac:
Using const where appropriate.   Typically we want to use const ref for args to getter functions
and to avoid unnecessary construction of temporary objects.

Moving typedef up front makes it easier for new developers to see the recommended use pattern.
i.e.
-               static std::shared_ptr<rogue::interfaces::memory::Variable> create (
+               static rogue::interfaces::memory::VariablePtr create (
We want end users to see the VariablePtr typdef when searching the docs for the create API's instead of the more dense std::shared_ptr<rogue::interfaces::memory::Variable>.

### DIFF
--- a/include/rogue/LibraryBase.h
+++ b/include/rogue/LibraryBase.h
@@ -67,8 +67,8 @@ namespace rogue {
          //! Get variable by name
          std::shared_ptr<rogue::interfaces::memory::Variable> getVariable(const std::string & name);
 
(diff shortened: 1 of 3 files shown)


slacrherbst:
Moving the typedef up front breaks doxygen linkages.

slacrherbst:
THe other changes are fine.

PR #689, opened 2020-07-02
Template friendly

bhill-slac:
I'm using a C++ template based approach on the EPICS driver so need each variable type to support a getValue() and setValue() call w/ the appropriate type in the args.


slacrherbst:
Approved but there are conflicts with the previous PR. Please resolve.

PR #690, opened 2020-07-02
Variable set log level

bhill-slac:
Allows developers to adjust logging for a specific variable and it's underlying blocks while running the app.


slacrherbst:
Can you add the following line to Variable.cpp so we can expose this to pyhon:

      .def("_setLogLevel",     &rim::Variable::setLogLevel)

PR #702, opened 2020-07-15
Add DumpVars button so users can initiate call to remoteVariableDump

bhill-slac:
If this isn't how you would add a button feel free to change.
Just need a way to initiate a dump once I have gui in desired state.


slacrherbst:
This already exists as a feature.

PR #705, opened 2020-07-15
Ensure values passed to get bytes are initialized

slacrherbst:


src/rogue/interfaces/memory/Block.cpp
```diff
@@ -532,6 +532,8 @@ void rim::Block::setPyFunc ( bp::object &value, rim::Variable *var ) {
 bp::object rim::Block::getPyFunc ( rim::Variable *var ) {
    uint8_t * getBuffer = (uint8_t *)malloc(var->byteSize_);
 
+   memset(getBuffer,0,var->byteSize_);
+
```
bengineerd, line 536: It probably doesn't make much difference, but isn't this basically what `calloc()` is for?

PR #709, opened 2020-07-21
Use PyDMLabels for read only variables

slacrherbst:


ruck314:
So much easier to read now:
image

ruck314 (approved):

PR #711, opened 2020-07-21
Writeable only arg for remote variable dump

bhill-slac:
Makes it easier to generate and compare config files.


slacrherbst:
A couple of notes.

First there is no need to add any of these items into the GUI. They will show up in the commands tree. 

These new command should be hidden by default and then unhidden by the cameralink specific root level.

I will create a new PR which includes the correct filtering type for these commands.

slacrherbst:
Changes included in #722

PR #712, opened 2020-07-22
Fix memory leak in epics3 valueSet call

slacrherbst:
It turns out there was a mistake in the epics3 valueSet call related to setting array values. This was causing a memory leak.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,12 +113,8 @@ jobs:
         # Run tests with coverage
         - cd $TRAVIS_BUILD_DIR
         - python3 -m pytest
-        #- python3 -m pytest --cov
 
       after_success:
-        # Upload coverage report
-        #- codecov
-        #- coverage report -m
         # Publish docs.
         - if [ "$TRAVIS_BRANCH" = "documentation" ]; then
            doctr deploy . --built-docs docs/build/html --deploy-branch-name gh-pages --branch-whitelist documentation;
--- a/include/rogue/protocols/epicsV3/Value.h
+++ b/include/rogue/protocols/epicsV3/Value.h
@@ -82,6 +82,8 @@ namespace rogue {
                //! Class creation
                Value ( std::string epicsName );
 
+               ~Value ();
+
                std::string epicsName();
 
                virtual bool valueSet();
--- a/pip_requirements.txt
+++ b/pip_requirements.txt
@@ -2,8 +2,6 @@ PyYAML
 jsonpickle
 parse
 click
-coverage
-codecov
 pytest
 pyepics
 doctr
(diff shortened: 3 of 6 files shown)


jesusvasquez333:
I think there are other memory leaks in the EPICS server code. I see calls to `new` without a later call to `delete`. 

For example here:
https://github.com/slaclab/rogue/blob/ESCRYODET-640/src/rogue/protocols/epicsV3/Value.cpp#L170

I think that code should look like this:
```c++
if (pValue_)
    delete pValue_;

pValue_ = new gddAtomic (gddAppType_value, epicsType_, 1u, count);
```

Similarly here as well:
https://github.com/slaclab/rogue/blob/ESCRYODET-640/src/rogue/protocols/epicsV3/Value.cpp#L179
https://github.com/slaclab/rogue/blob/ESCRYODET-640/src/rogue/protocols/epicsV3/Value.cpp#L263
https://github.com/slaclab/rogue/blob/ESCRYODET-640/src/rogue/protocols/epicsV3/Value.cpp#L272

Additionally, the destructor of the class should do something like this as well:
```c++
rpe::Value::~Value()
{
    if (pValue_)
        delete pValue_;

    if (units_)
        delete units_;
}
```

Although a better and safer approach will be to use smart pointers. Change all the data types from `gdd*` to `std::shared_ptr<gdd>`. Then we don't need to worry to explicitly call `delete`.

slacrherbst:
In general I agree with you idea of using smart pointers, but the pcas code is somewhat limiting there. 

The gddAtomic class uses its own cleanup mechanisms. They have a dereference calls which is used to destroy the object when the reference count drops to zero. 

The pValue_ stays active for most of the life of the entity. When a new value is passed the previous value is de-referenced before the new object is assigned. I can't explicitly delete it as the core epics code may still be accessing it. here is an example of re-assigning and de-referencing pValue_:

https://github.com/slaclab/rogue/blob/f16a7508250bceed3c51929616a1d26414ed433d/src/rogue/protocols/epicsV3/Value.cpp#L271-L272

I walked through the code last night and I am confident that we don't have any other runtime memory leaks. I do agree that the current code may not properly clean up all memory allocations at deconstruction. That is less of a concern as this object stays persistent during the lifetime of the rogue entity. But I will put some time into cleaning that up as well.

jesusvasquez333:
@slacrherbst Ok, thanks for the clarification. I didn't notice the calls to `pValue_->unreference();`.

slacrherbst:
I added cleanup of the remaining values in object destructor:

https://github.com/slaclab/rogue/pull/712/commits/887db97d037cb4c9e856281bb4fb179673d800e9

PR #715, opened 2020-07-23
Auto expand small lists

slacrherbst:
This adds are arg to the VariableList widget 'maxListExpand' which defines the max number of entries a list can have and be auto expanded. This arg is  also available at the runPyDM() method.


ruck314 (approved):

ruck314:
image
Looks good.  I verified that I can control the expanding.  `maxListExpand = 1` will not expand and `maxListExpand >1` will expand `DataCnt[1:0]`

PR #719, opened 2020-08-10
Add support for list variables in both Local and Remote variables

slacrherbst:
This PR adds support for lists contained in variables. This was partially supported previously in LocalVariables but not supported in RemoteVariables. 

With this PR each element of the list is processed using the disp attributes independently for both Local and Remote variables.

A list based remote variable can be created by adding the following three attributes during the remote variable creation:

numValues: the number of values in the list
valueBits: the number of bits for each value
valueStride: the spacing between each of the variables.

the bitOffset and bitSize attributes are still used. they must only contain one value each, as sparse registers are not support when using list variables. The bitSize must be equal to the numValues * valueStride. 

Values set and get can be used with an entire list through the classic set(), get(), value(), getDisp(), valueDisp, setDisp() calls. Individual values in the list can be set by adding the index=n arg to the above calls. This new arg is also passed through the set and get calls in a LinkedVariable if the user wishes to support it. 

Calling set() and get() without index results in a read/write of the full block range. Adding index= limits the read/write to the memory range being accessed. 

The config file also supports sliced assignments:
ListVar[1:2] = [10]
ListVar[1] = 10
ListVar[1:5] = [1,2,3,4]

List variable support is also included in the c++ layer access APIs.


include/rogue/LibraryBase.h
```diff
@@ -89,7 +89,7 @@ namespace rogue {
          uint64_t getFieldUInt64(std::map<std::string, std::string> fields, std::string name);
```
ruck314, line 89: Should the other types (e.g. getFieldUint64) have a `bool def=false` argument?

python/pyrogue/_Block.py
```diff
@@ -15,6 +15,14 @@
 import pyrogue as pr
 
 
+def startTransaction(block, *, type, forceWr=False, checkEach=False, variable=None, index=-1, **kwargs):
```
ruck314, line 18: Should there be a print statement about how `startTransaction` is depreciated and should use `_startTransaction` instead?

python/pyrogue/_Block.py
```diff
@@ -15,6 +15,14 @@
...
+    block._startTransaction(type, forceWr, checkEach, variable, index)
+
+
+def checkTransaction(block, **kwargs):
```
ruck314, line 22: Should there be a print statement about how checkTransaction is depreciated and should use _checkTransaction instead?

include/rogue/LibraryBase.h
slacrherbst: Maybe eventually as required. In reality this may just be a short term stop gap to avoid issues with .h files generated with the old API.

python/pyrogue/_Block.py
slacrherbst: Designs will need to use the startTransaction and checkTransaction helper functions in the future. I can add a note to the top of block.py.

python/pyrogue/_Block.py
slacrherbst: Same as above. The idea is the helper function will absorb and future proof the function args.

PR #722, opened 2020-08-13
Changes from Bruce's fork

slacrherbst:


bhill-slac:
Changes look good to me.    As I recall I needed the cmake changes to ensure I got full rpath info in the installed shared libs.   That can be helpful when someone submits a bug report re an IOC and avods issues w/ LD_LIBRARY_PATH.    Control systems engineers avoid updating IOCs unless they are fixing a problem, so we'll be deploying in an environment w/ potentially many different versions of the rogue libraries in use by different IOCs.

PR #733, opened 2020-09-15
Break large array into sections on gui

slacrherbst:
Avoid building arrays larger than 100 entries on expansion. This also groups node fetches into bursts which reduces the time to interact with the server. 

This also fixes an epics bug already fixed in Rogue 5.


slacrherbst:
This PR is ready for review.

PR #736, opened 2020-09-16
Use native pickling instead of JSON for zmq interface

slacrherbst:
This change uses native python pickling instead of json encoding for sending data between the zmq client and server. 

This also uses raw bytes for data transfer between the client and server instead of passing and decoding strings.

The C++ raw client interface is still supported on a third server port.

This has resulted in a 50% speedup of zmq based transactions which should make the GUI more responsive. The Rogue zmq remains faster than the epics3 interface:

For a list transfer:
   Epics list caget Rate = 1132.6 hz
   Epics list caput Rate = 1218.6 hz
   Epics list pv.get Rate = 1145.9 hz
   Epics list pv.put Rate = 1261.5 hz
   zmq list get Rate = 3159.4 hz
   zmq list set Rate = 3048.9 hz

For a single value transfer:
   Epics caget Rate = 1668.3 hz
   Epics caput Rate = 2379.5 hz
   Epics pv.get Rate = 2105.4 hz
   Epics pv.put Rate = 2457.7 hz
   zmq get Rate = 4318.7 hz
   zmq set Rate = 4028.2 hz

### DIFF
--- a/tests/epics_client_testing.py
+++ b/tests/epics_client_testing.py
@@ -1,0 +1,184 @@
+
+import epics
+import time
+import pyrogue.interfaces
+
+#class EpicsInterface(object):
+#
+#    def __init__(self):
+#        self._epicsError = None
+#        epics.ca.replace_printf_handler(self._printHandle)
+#
+#    def _printHandle(self, *argv, **kwargs):
+#        print("Callback")
+#        print(argv)
+#        print(kwargs)
+#        print("Done")
+#        #print(f"Got {argv} {kwargs}")
+#        #self._epicsError = msg
+#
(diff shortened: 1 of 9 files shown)


bengineerd:
I assume this doesn't break backward compatibility of the ZMQ protocol?

slacrherbst:
It does. The client and server versions will have to match.

slacrherbst:
This PR is now ready and the old raw CPP interface has been restored.

tests/epics_client_testing.py
```diff
@@ -0,0 +1,184 @@
+
```
ruck314, line 1: missing SLAC licensing header

PR #738, opened 2020-09-18
Improve handling of lists and numpy arrays in epics3 & epics4

slacrherbst:
This PR improves the general handling of lists and numpy arrays, particularly when using the epics3 & epics4 protocols. This PR also extracts the numpy type and displays the typeStr properly to indicate the value is a numpy array and the underlying value type, for example:

Int32[np] 

where [np] indicates a numpy array type vs [] or a standard list. In epics3 numpy arrays are converted to strings.


slacrherbst:
No comments on the PR. I will go ahead and merge.

PR #742, opened 2020-09-25
Update MemoryDevice to use its own transaction logic

slacrherbst:
This PR installs a local txnChunker method in the MemoryDevice class in preparation for the rawWrite and rawRead methods to be deprecated from Device. MemoryDevice now has its own lock object and will store size in a local variable which will be preserved when size goes away from the Device class. 

We will keep MemoryDevice around for times when the accessed memory is too large for local shadow memory.

### DIFF
--- a/python/pyrogue/_Memory.py
+++ b/python/pyrogue/_Memory.py
@@ -40,7 +40,6 @@ class MemoryDevice(pr.Device):
         )
 
         self._size    = size
-        self._lockCnt = 0
         self._txnLock = threading.RLock()
 
         if isinstance(base, pr.Model):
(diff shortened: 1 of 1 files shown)


slacrherbst:
I added the size to the write call and updated verifyBlocks to check the state of the _verify flag.

ruck314:
@slacrherbst Should there be an update to the _rawWrite/_rawRead that prints out that these two function are being deprecated for the 1st time (or everytime) that it is used?

ruck314 (approved):

slacrherbst:
I am going to add that to a future version.

PR #744, opened 2020-10-06
Update _DataReceiver.py

ruck314:
### Description
- Adding missing `__rshift__` and `__lshift__`

### DIFF
--- a/python/pyrogue/_DataReceiver.py
+++ b/python/pyrogue/_DataReceiver.py
@@ -101,12 +101,20 @@ class DataReceiver(pr.Device,ris.Slave):
         self.Data.set(dat,write=True)
         self.Updated.set(True,write=True)
 
-
     def _start(self):
         super()._start()
         self._rxEnable = True
 
-
     def _stop(self):
         self._rxEnable = False
         super()._stop()
+
+    # source >> destination
+    def __rshift__(self,other):
+        pr.streamConnect(self,other)
+        return other
+
+    # destination << source
+    def __lshift__(self,other):
+        pr.streamConnect(other,self)
+        return other


slacrherbst:
Is this necessary? It should pick these up from the Slave base class.

ruck314:
This is why you can't use `<<` or `>>` in wave8 and had to used the `pyrogue.streamConnect()` instead:
https://github.com/slaclab/wave8/blob/master/firmware/python/wave8/Top.py#L233

I have ran into this before when there is multiple parents classes for a single child class:
https://github.com/slaclab/sparkPix-S/blob/master/firmware/python/spark_pix_s/_SimPatternGen.py#L166

slacrherbst:
Interesting. This appears to work in the test script. Have we double checked that this fails:

https://github.com/slaclab/rogue/blob/cad522dece68cc9889d5541f2e53046bf3a6bccd/tests/run_server.py#L110-L113

slacrherbst:
> I have ran into this before when there is multiple parents classes for a single child class:
> https://github.com/slaclab/sparkPix-S/blob/master/firmware/python/spark_pix_s/_SimPatternGen.py#L166

I think that was required because of base class confusion.

ruck314:
`pyrogue.streamConnect(self.epicsUb,d)`  = works
`self.epicsUb >> d` = works
`d << self.epicsUb` = broken
```
Rogue/pyrogue version v5.4.0-27-gfafb7a44. https://github.com/slaclab/rogue
  File "scripts/wave8DAQ.py", line 147, in <module>
    timeout     = 5.0,
  File "/afs/slac.stanford.edu/u/re/ruckman/projects/wave8/firmware/python/wave8/Top.py", line 235, in __init__
    d << self.epicsUb                       # Broken
rogue.GeneralError: memory::Slave::lshiftPy: General Error: Attempt to use << operator with incompatible memory master

**Comment by slacrherbst** [comment]
I just tested the RawDataReceiver.py class in the test script. What error do you get? I wonder if it is related to the other class.

**Comment by ruck314** [comment]
```
Rogue/pyrogue version v5.4.0-27-gfafb7a44. https://github.com/slaclab/rogue
  File "scripts/wave8DAQ.py", line 147, in <module>
    timeout     = 5.0,
  File "/afs/slac.stanford.edu/u/re/ruckman/projects/wave8/firmware/python/wave8/Top.py", line 235, in __init__
    d << self.epicsUb                       # Broken
rogue.GeneralError: memory::Slave::lshiftPy: General Error: Attempt to use << operator with incompatible memory master
```

slacrherbst:
Ok. perhaps the interrogation is bad. I will test that.

slacrherbst:
Ok I see now. It appears the problem is related to the CPP implementation of the lshift code in the slave class (as called in the DataReceiver class). It appears it cannot properly extract the master class from a class that inherits from both a master and a slave (unbatcher/splitter). Let me look into that.

slacrherbst:
I am going to approve and merge this but it is a bandaid fix. The error appears to be in the way boost::python::extract works. It does not properly extract the Master stream class object in the __lshift__ call. This fix can be applied to python subclasses but it does not fix cases where you try to connect a pure C++ slave to a C++ master/slave.

PR #765, opened 2021-02-23
Limit size of running SystemLog variable

slacrherbst:
This limits the number of entries in the SystemLog variable to a configurable number of entries (default=1000). The number of entries can be changed using the MaxLog parameters in the Root class. 

This also limits the number of displayed entries in the PyDm console to the last 20.


jesusvasquez333 (approved):

jesusvasquez333:
See test results here: https://jira.slac.stanford.edu/browse/ESCRYODET-728

PR #770, opened 2021-03-03
Allow the user to use StreamWriter with custom writers

slacrherbst:


bengineerd (approved):
I modified `LegacyStreamWriter` to use this new feature.

Do any other `DataWriter` classes exist yet?

PR #774, opened 2021-03-15
Add default value for index parameter in get and set

bengineerd:
This PR adds a default value of -1 to the `index` parameter of `Block.set()` and `Block.get()`.

The `_iadd()`, `_isub()` and friends all call both `get()` and `set()` without an index, so it seems easiest to provide a default.

### DIFF
--- a/python/pyrogue/_Block.py
+++ b/python/pyrogue/_Block.py
@@ -89,7 +89,7 @@ class LocalBlock(object):
     def variables(self):
         return self._variables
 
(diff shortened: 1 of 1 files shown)


bengineerd:
After some thought, I think it's fine for `LocalBlock` to allow a default index even though C++ Blocks do not.

Can we merge this and cut a release for so the KPIX software works?

slacrherbst (approved):

PR #779, opened 2021-04-20
Fix syntax error

bengineerd:
Fix syntax error in `LegacyStreamWriter` init.


bengineerd:
This was already pushed directly to the master branch.

PR #781, opened 2021-04-30
Allow non-blocking get() and set()

bengineerd:
Several parameters have been added to Variable.get(), Variable.set(), and related methods.

For all methods, all parameters other than `value` are now keyword only arguments, which corresponds to how we generally call them. This will also avoid confusing calls like `var.set(value, True, False, True)`

The `check` argument to `set()` and `get()` is used to skip the internal `Device.checkBlocks()` call. This effectively creates a non-blocking call. This can be useful when overriding `writeBlocks()`. Some devices require a "load" register to be toggled once the full configuration has been written. We typically solve this by calling `self.load.set(1)` within `writeBlocks()`. But that `set()` call then blocks during a `LoadConfig()` operation that doesn't really need to block. 

The non-blocking `Variable.get()` call allows multiple reads to be queued by a script, with Variable listeners being informed of the updates once they become available.

The `verify` argument allows the automatic verify operation after a write to be skipped. This may be useful when scripting or within `writeBlocks()`.

### DIFF
--- a/python/pyrogue/_Variable.py
+++ b/python/pyrogue/_Variable.py
@@ -336,7 +336,7 @@ class BaseVariable(pr.Node):
             self.__functions.remove(listener)
 
     @pr.expose
-    def set(self, value, write=True, index=-1):
+    def set(self, value, *, index=-1, write=True, verify=True, check=True):
         """
         Set the value and write to hardware if applicable
         Writes to hardware are blocking. An error will result in a logged exception.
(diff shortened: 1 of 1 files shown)


slacrherbst:
Do we still want to look into this or close this PR?

bengineerd:
I think this is worth keeping. It's fully backward compatible, and adds some versatility to handle corner cases in register transaction sequencing. I need to reverify that it works as expected. I can test that this week.

PR #782, opened 2021-04-30
Allow optional name to be passed to many interface types

bengineerd:


slacrherbst:
This appears to create a lot of issues. We will shelve for now.

PR #783, opened 2021-05-03
Fix stride bug in txnChunker

bengineerd:
`MemoryDevice._txnChunker()` and `Device._rawTxnChunker()` were not calculating the stride correctly when `wordBitSize < 32` and data was passed in as a list of values.


ruck314:
@bengineerd Can you add a script in `rogue/tests` for testing the corner cases for this bug fix?

slacrherbst (approved):

PR #796, opened 2021-09-02
Defer pydm import

bhill-slac:
Needed when we use rogue on a linuxRT system as we don't have QT support there.
However, we still need to use the non-gui parts of rogue w/o triggering an exception.

### DIFF
--- a/python/pyrogue/__main__.py
+++ b/python/pyrogue/__main__.py
@@ -11,8 +11,6 @@
 import argparse
 import pyrogue
 import pyrogue.interfaces
-import pyrogue.gui
-import pyrogue.pydm
 import time
 import os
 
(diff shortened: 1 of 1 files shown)


slacrherbst:
This PR breaks the ability to run the gui from the pydm module. This can be fixed by adding:
        import pyrogue.pydm

to line 67:

# GUI Client
if args.cmd == 'gui':
    import pyrogue.pydm
    pyrogue.pydm.runPyDM(serverList=args.server,ui=args.ui)

bhill-slac:
I don't understand your comment as I already had an import pyrogue.pydm there, along w/ import pyrogue.gui.   Maybe the latter wasn't really needed at all?    Have posted my  branch to slaclab/rogue if you want to fix it the way you want.

bhill-slac:
It would really help to get this into an /afs/slac/g/reseng conda release.   Currently I need to hack my conda rogue/__main__.py in PROD to get through conda activation.

slacrherbst (approved):

PR #805, opened 2021-09-27
Remove epics3/PCAS from rogue

slacrherbst:
This PR removes the epics3 PCAS layer from Rogue. This is currently used in two known projects.

- SMURF  which will move to zmq if rogue is ever upgraded
- HPS which may need to look for an alternative epics gateway in the future.


bhill-slac:
I don't understand why you're removing EPICS v3.    AFAIK DAQ still uses this for testing and development when they don't have ready access to an EPICS IOC for the device.

bhill-slac:
I've been using EPICS alias'es to ensure that the rogue PV paths still work to make it easy for DAQ to work w/ either an EPICS IOC or the rogue python w/ EPICS v3.

slacrherbst:
> I don't understand why you're removing EPICS v3. AFAIK DAQ still uses this for testing and development when they don't have ready access to an EPICS IOC for the device.

I checked with Matt and Chris and they state they are only using p4p (EPICS 4) for their interface.

slacrherbst:
> I've been using EPICS alias'es to ensure that the rogue PV paths still work to make it easy for DAQ to work w/ either an EPICS IOC or the rogue python w/ EPICS v3.

Can you do this with the epics4 server and an epics4 client?

slacrherbst:
If truly necessary we could keep the epicsCA support, but I would need to drop list support, instead passing them as strings.

slacrherbst:
Closing this PR in favor of a more staged deprecation.

PR #806, opened 2021-10-01
Change the way pydm handles commands

slacrherbst:
This PR changes the way pydm deals with commands.

First if a command takes an arg we simply provide an input box for that arg. Updates to the input box will execute the command without the need to hit a separate exec button. If a command does not take an arg the exec button is provided. 

Second, this PR combines the variable and command trees into a single tree for simplicity.

Also the last passed arg is cached in a command so that the input box is updated on other connected guis.


slacrherbst:
I updated the debug tree to use a separate column for commands and variables.

PR #808, opened 2021-10-01
Deprecate epics v3

slacrherbst:
This PR begins the process of deprecating epics v3. The primary change is to remove list support from the PCAS server as it was broken with the changes to numpy arrays. The EPICS library is no longer included by default. To build with epics the user must pass the -DDO_EPICS arg to cmake.


slacrherbst:
Another note. I am going to properly handle numpy arrays as vectors in the up_update PR.

slacrherbst:
@bhill-slac Are you good with the end result?

slacrherbst:
@bhill-slac I will set up the github actions to generate a rogue-epics3 package for anaconda.

PR #825, opened 2022-01-11
Update _Memory.py

ruck314:
### Description
- Adding `update()` function for users to do a "bulk" remote memory update
- It is best to prevent users from having to use writeBlocks(), verifyBlocks() and checkBlocks() in their application code


bengineerd:
Can you point me to some application code that calls this new `update()` method?

PR #826, opened 2022-01-11
ESROGUE-532: Bug fix for `Bad access to shared pointer when creating UDP client and server`

mskoufis:
### Description
This fix bypasses the bad shared_ptr issue by creating a Pool object and then calling its member functions, instead of making direct calls to the functions in the un-instantiated class.  It seemed to me that's why shared_from_this() was failing before, as this call will work only with an object already owned by a shared_ptr.

### DIFF
--- a/include/rogue/interfaces/stream/Pool.h
+++ b/include/rogue/interfaces/stream/Pool.h
@@ -73,6 +73,9 @@ namespace rogue {
 
             public:
 
+               //! Class creation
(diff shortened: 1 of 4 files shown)


bengineerd:
@ruck314 and @slacrherbst Does this fix work for you?

bengineerd (approved):
Solution looks reasonable to me, but I'm no expert on shared pointers. Would like @slacrherbst opinion.

mskoufis:
Closing this pull request since the fix is not correct.

PR #828, opened 2022-01-14
ESROGUE-539: Add accessor to set the SRP hardware timeout

mskoufis:


ruck314:
@mskoufis How do I access `setHardwareTimeout()` from python?

mskoufis:
@ruck314 I haven't implemented a python wrapper for this functionality.  Perhaps I should look at that.

bengineerd:
The timeout value should be an instance variable of the SrpV3 class. The set function sets the instance variable and that gets used when building every header.

bengineerd (commented):
The timeout value is stored in bits 31:24 of `header[0]`. You'll need to shift and mask `timeout` as you assign it into `header[0]`.

Also the style we use for instance variables is to end with `_`, so the variable should be named `timeout_`.

mskoufis:
Okay, I had assumed that the shifted value would be provided to this function.  I'll work on that.

bengineerd:
Ok, now let's expose `setHardwareTimeout()` to python. 
Master.cpp provides a pretty good example of how to do it.
https://github.com/slaclab/rogue/blob/main/src/rogue/interfaces/memory/Master.cpp#L46-L71

bengineerd:
Also, I know it's a simple one-liner, but I would prefer that the `setHardwareTimeout()` implementation be in the .cpp file for style consistency.

mskoufis:
Okay, no problem.

src/rogue/protocols/srp/SrpV3.cpp
```diff
@@ -55,9 +55,9 @@ rps::SrpV3Ptr rps::SrpV3::create () {
 //! Setup class in python
 void rps::SrpV3::setup_python() {
 #ifndef NO_PYTHON
-
-   bp::class_<rps::SrpV3, rps::SrpV3Ptr, bp::bases<ris::Master,ris::Slave,rim::Slave>,boost::noncopyable >("SrpV3",bp::init<>());
-
+   bp::class_<rps::SrpV3, rps::SrpV3Ptr, bp::bases<ris::Master,ris::Slave,rim::Slave>,boost::noncopyable >("SrpV3",bp::init<>())
+      .def("_setHardwareTimeout",           &rps::SrpV3::setHardwareTimeout)
+   ;
```
ruck314, line 60: Can you make the code and semi-colon on the "same line"

ruck314 (approved):

bengineerd (approved):

PR #829, opened 2022-01-14
ESROGUE-533: Shutdown Rogue if DmaDriver version mismatch

mskoufis:


ruck314:
@mskoufis I think you might need to add this to the PgpCard.cpp as well:
https://github.com/slaclab/rogue/blob/master/src/rogue/hardware/pgp/PgpCard.cpp#L69

mskoufis:
@ruck314 Sounds good.  Will do.

ruck314:
@mskoufis I would also update the comment to mention that the new version is a 64-bit address API
```
To use later versions (64-bit address API), you will need to upgrade ....
```

ruck314:
@mskoufis Change this statement `v5.12.2 (or later ... or next release)` to `v5.13.0 (or later)`.  This will assume the next rogue tag release is v.5.13.0

ruck314 (approved):

PR #830, opened 2022-01-15
ESROGUE-518: Load and auto-populate registers based on Xilinx *_hw.h header

mskoufis:


ruck314:
@mskoufis Does this support the ability to point to a HLS .zip file and have it automatically find teh hw_.h file?  Else the user would need to decompress and hunt for it manually everytime

mskoufis:
@ruck314 No it does not.  It's expecting that the path to the *_hw.h will be provided.

mskoufis:
I made the change to extract the register interface *_hw.h file from the zip file directly.  I assumed there will be only one such file in that archive.

slac-ncfronk:
In my experience with hls, the _hw.h file can provide memory blocks of arbitrary size and the remote_variable class cannot handle >64 bit widths. Is that a problem for this utility?

mskoufis:
@slac-ncfronk The tool doesn't do any checks in order to approve or reject bit widths or other parameters it finds in *_hw.h.  Also note that this utility does not provide a fully automated solution for populating the remote variables from the register interface file.  The user or developer has to modify the output so that it's usable.  In that sense, even if some of the bit widths exceed the value of 64, they can be changed manually during the second (manual) pass.  @ruck314 @slacrherbst Is that acceptable?

bengineerd:
Let's try to add the bit width check.
The pseudo code is something like this:
```python
if valueBits > 64:
  for i in range(numValues):
    pr.add(pr.RemoteVariable(
      name = f'{var_name}[{i}]',
      ...
```
Basically instantiate a RemoteVariable for each "value" in the array.

python/pyrogue/utilities/hls/_RegInterfParser.py
```diff
@@ -0,0 +1,265 @@
...
+            var = MemoryDevice(name=dname, offset=addr, size=depth, wordBitSize=width)
+            params.append(var)
+        elif vartype is RemoteCommand.__name__:
+            var = RemoteCommand(name=dname, description='remote command', offset=addr, bitSize=width)
+            params.append(var)
+
+    return params
+
+def write(parameters):
+    fhand = open('application.py', 'w+')
```
bengineerd, line 176: Should use a context manager here.

```python
with open('application.py', 'w+') as fhand:
```

mskoufis:
I think this is done, unless someone has anything else to add.

PR #831, opened 2022-01-25
ESROGUE-540: PyDM FileWriter - Prevent users from updating the "Data File" string while the file is opened

mskoufis:


ruck314 (approved):

bengineerd:
I've tested this and it looks like it works.

While we're at it with this, I think the buttons should be disabled out based on file open status as well.
File not open -> "Close" button disabled
File open -> "Browse", "Auto Name", "Open" disabled

Thoughts @ruck314?

@mskoufis Is this hard to do?

mskoufis:
@bengineerd Im sure it's doable if you think it's nice to have.

bengineerd:
@mskoufis Ok, go ahead and implement it.

mskoufis:
@bengineerd Done.  Seems to work as intended, unless I missed something.

bengineerd:
Can you disable the "Close" button on start-up? It doesn't get disabled until a file has been opened and then closed.

mskoufis:
I believe this is ready for integration.

PR #832, opened 2022-01-25
ESROGUE-548: Make DataReceiver.RxEnable work properly

bengineerd:
The `RxEnable` `LocalVariable` is now used to control the behavior of `_acceptFrame()`.

Previously it was controlled by an internal instance variable `_rxEnable`, and not accessible via the Rogue API. The `RxEnable` rogue variable was confusingly left unused. This is now fixed.


bengineerd:
@ruck314 I don't have any active projects that use `DataReceiver`. Are you able to verify that this fix works the way you expect?

ruck314 (approved):

Issue #823, opened 2021-12-07
Bad access to shared pointer when creating UDP client and server 


maTizK:
Constructor of `rogue::protocols::udp::Client` and `rogue::protocols::udp::Server` spawns a `runThread`. Inside this thread a `shared_from_this()` is called, that returns a shared pointer of the constructed object. 
In the case where the object is already constructed, e.g. shared pointer is not weak it works fine.
However, if the spawned thread `runThread` tries to generate a shared pointer before the object is constructed it throws `std::bad_weak_ptr`. 

Now, it seems that `runThread` of `Client` and `Server` has a `usleep(1000);` to overcome this problem. However, this is not always sufficient. The `runThread` shall wait until the object is created, i.e. shared pointer was created out of it. 

For example with use of mutex
```cpp
//! Class creation
rpu::ServerPtr rpu::Server::create(uint16_t port, bool jumbo)
{
   rpu::ServerPtr r = std::make_shared<rpu::Server>(port, jumbo);
   r->constructed();
   return (r);
}

//! Set object is constructed
void rpu::Server::constructed()
{
   mutex_.unlock();
}

//! Run thread
void rpu::Server::runThread()
{
   ris::BufferPtr buff;
   ris::FramePtr frame;
   fd_set fds;
   int32_t res;
   struct timeval tout;
   struct sockaddr_in tmpAddr;
   uint32_t tmpLen;
   uint32_t avail;

   udpLog_->logThreadId();
   std::lock_guard<std::recursive_mutex> lk(mutex_); // wait for construction to end

   // Preallocate frame
   frame = ris::Pool::acceptReq(maxPayload(), false);

   while (threadEn_)
   {
```

slacrherbst:
Jesus and I could not figure out a solution to this.  We have this problem in numerous places:


https://github.com/slaclab/rogue/commit/2b4358cf5107b7fbcdf272b6c71c8ab6ef80b1a0

This appears to be a clever solution to this.

maTizK:
Maybe this example helps to solve the issue since locking mutex introduces a new possible deadlock. In this example it's required to call `destroy()` or in your case `stop()`. With this you avoid any undefined behaviour if destructor gets called in between construction and spawning a thread. 

```cpp
#include <chrono>
#include <future>
#include <iostream>
#include <memory>
#include <string>
#include <thread>
#include <unistd.h>
#include <vector>
#include <date/date.h>

struct Logger
{

  template <typename Arg0>
[...]

PR #836, opened 2022-02-02
ESROGUE-532: Bug fix for bad weak_ptr when calling shared_from_this()

mskoufis:
### Description
The call to shared_from_this() from within a thread would fail due to the constructor launching that thread not completing in time.

### JIRA:
[ESROGUE-532](https://jira.slac.stanford.edu/browse/ESROGUE-532)

### Issues
#823

### DIFF
--- a/include/rogue/hardware/axi/AxiStreamDma.h
+++ b/include/rogue/hardware/axi/AxiStreamDma.h
@@ -68,7 +68,7 @@ namespace rogue {
                std::shared_ptr<rogue::Logging> log_;
 
                //! Thread background
-               void runThread();
+               void runThread(std::weak_ptr<int>);
 
                //! Enable zero copy
                bool zeroCopyEn_;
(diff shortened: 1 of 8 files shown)


mskoufis:
@slacrherbst Can you review this when you get a chance?  I'd like to close out this week, if at all possible.

slacrherbst:
The idea is that the weak pointer expires when the constructor exist?

mskoufis:
Yes indeed

slacrherbst:
Clever. We should also apply this fix to other classes that start a thread with a shared from this call in run thread.

mskoufis:
You mean as part of this fix or as a separate ticket?

slacrherbst:
Good question either works. Here is an example of another place we added a sleep:

https://github.com/slaclab/rogue/blob/4be0e9a4d17bdd3987a268f54ad195ee1093190d/src/rogue/hardware/axi/AxiStreamDma.cpp#L371

mskoufis:
Okay I see.  I'll search for usleep then and see what else comes up.

bengineerd (approved):

PR #839, opened 2022-02-03
ESROGUE-545: Disable Load and Save buttons in GUI when file text fields are blank

mskoufis:
### Description
Disable `Load Config`, `Save Config` and `Save State` buttons when the file path boxes are empty.

### JIRA
https://jira.slac.stanford.edu/browse/ESROGUE-545

### DIFF
--- a/python/pyrogue/pydm/widgets/root_control.py
+++ b/python/pyrogue/pydm/widgets/root_control.py
@@ -190,7 +190,7 @@ class RootControl(PyDMFrame):
         if isinstance(saveFile,tuple):
             saveFile = saveFile[0]
 
-        if saveFile != '':
+        if len(saveFile) != 0:
             self._saveConfigValue.setText(saveFile)
             self._saveConfigCmd.setEnabled(True)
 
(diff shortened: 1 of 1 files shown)


ruck314:
@mskoufis I am unable to open/define the file that I want to create:
image

mskoufis:
@ruck314 Is that when you do Load Config?  It sounds like it's expecting an existing configuration file?

mskoufis:
@ruck314 I incorporated your suggestions and it's ready to test again.

ruck314:
@mskoufis One issue that might come up is that you have to use the browser to enable the button.  So a user would not be able to cut/paste the path into the test box.  Can you update the buttons to enable if text is typed into the text box?
image

mskoufis:
@ruck314 I already did that for load config.  Can you test?  I will implement it for the other two cases (i.e. save config, save state).

ruck314:
@mskoufis Behavior for load config.  Go ahead an apply to other two buttons

mskoufis:
@ruck314 It should be done and ready to go.

ruck314 (approved):
Looks good now.  Ready to merge

PR #842, opened 2022-02-05
Removing the obsolete,legacy PyQT GUI from mainline rogue

ruck314:
### Description
- After a few years of warning about depreciating this GUI, it is time to get rid of it


ruck314:
Ok, then we hold off on this until both all EPIX migrate to rogue v5 + PyDM GUI

PR #844, opened 2022-02-07
ESROGUE-526: Use portable format specifiers when printing inttypes

mskoufis:
### Description

Replace all `"%i"` and related format specifiers with portable equivalents specified in `<inttypes.h>`.

### JIRA

[ESROGUE 526](https://jira.slac.stanford.edu/browse/ESROGUE-526)

### DIFF
--- a/include/rogue/interfaces/stream/FrameAccessor.h
+++ b/include/rogue/interfaces/stream/FrameAccessor.h
@@ -20,6 +20,7 @@
 #ifndef __ROGUE_INTERFACES_STREAM_FRAME_ACCESSOR_H__
 #define __ROGUE_INTERFACES_STREAM_FRAME_ACCESSOR_H__
 #include <stdint.h>
+#include <inttypes.h>
 #include <memory>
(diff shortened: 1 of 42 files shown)


mskoufis:
This one passes the CI tests but I haven't observed the logging output.  Not sure what is the best way to do this.

ruck314 (approved):

bengineerd:
Please change these in any exception throws as well.

I did a grep on the code and there are still a lot of `"%i"` and `"%x"` instances still in there.

mskoufis:
Okay.  I only updated the logging.

mskoufis:
@bengineerd Thanks for finding the %i instances in the header file.  I was only looking at the source files not the headers.  Also, I didn't change the sscanf lines in Value.cpp because I wasn't sure if that would change the expected format string and cause any problems.

bengineerd:
@mskoufis 
I think the `sscanf` calls are ok, but they should be SCNu32.
I'm not sure what the right way is to format `struct timeval` members. But what we have is probably ok.

PR #845, opened 2022-02-10
Write only the stale part of a Block

bengineerd:
This fixes a bug where writing only part of an array RemoteVariable caused the whole array to be written anyway

### Description

A bug was causing an entire RemoteVariable array to be written to hardware even if only a small portion of it had been updated with `RemoteVariable.set()`. This has been fixed.

### Details

`Block.intStartTransaction()` now decides which bytes to include in a write transaction based on the stale state.

### DIFF
--- a/src/rogue/interfaces/memory/Block.cpp
+++ b/src/rogue/interfaces/memory/Block.cpp
@@ -17,6 +17,7 @@
  * ----------------------------------------------------------------------------
 **/
 #include <rogue/interfaces/memory/Block.h>
+#include <rogue/interfaces/memory/Slave.h>
 #include <rogue/interfaces/memory/Variable.h>
 #include <rogue/interfaces/memory/Transaction.h>
 #include <rogue/interfaces/memory/Constants.h>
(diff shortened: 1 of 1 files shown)


src/rogue/interfaces/memory/Block.cpp
```diff
@@ -190,21 +192,28 @@ void rim::Block::intStartTransaction(uint32_t type, bool forceWr, bool check, ri
...
+             }
+         }
+         else {
 
-          if ( type == rim::Write || type == rim::Post ) {
+             lowByte = var->staleLowByte_;
+             highByte = var->staleHighByte_;
+             // Catch case where fewer stale bytes than min access or non-aligned
+             if (lowByte % minAccess != 0) lowByte -= lowByte % minAccess;
+             if ((highByte+1) % minAccess != 0) highByte += minAccess - ((highByte+1) % minAccess);
```
bengineerd, line 216: @slacrherbst This part here feels like a hack. Maybe `staleLowByte_` and `staleHighByte_` should obey the min access requirements at the time they are marked?

PR #847, opened 2022-02-18
Add value and valueDisp to node information

mskoufis:
Added val (i.e. value) and valDisp (i.e. valueDisp) to the node information list for a variable.

### Description

Since there were already accessor methods (without the @property decorator) used to get value and valueDisp and because changing those kept breaking the code, I decided (at least temporarily) to create new accessors in _Variable.py to get value and valueDisp that I called val and valDisp respectively.

### JIRA

ESROGUE-506

### DIFF
--- a/python/pyrogue/_Variable.py
+++ b/python/pyrogue/_Variable.py
@@ -380,6 +380,11 @@ class BaseVariable(pr.Node):
         return VariableValue(self,read=read)
 
     @pr.expose
+    @property
+    def val(self):
+        return self.value
+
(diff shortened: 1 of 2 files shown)


ruck314 (approved):

slacrherbst:
I don't think I agree with this

mskoufis:
@slacrherbst How do I modify the existing value() and valueDisp() to do what we want it?

bengineerd:
Why was this merged?

PR #850, opened 2022-02-22
Add `running` flag to `Device` to check whether `_start()` has been called

bengineerd:
### Description

The `Device.running` property is useful inside Device methods that access Variables but might fire before `Device._start()` has been called.

One such example is in `DataReceiver` (or any Device that is also a stream receiver), where `_acceptFrame` might be called before the Device tree has been started.

`DataReceiver` now also allows an `enableOnStart` instantiation parameter to allow `RxEnable` to be optionally set True once the `Device` has been started.

### JIRA

The need for the `running` flag was noticed after implementing https://jira.slac.stanford.edu/browse/ESROGUE-548 - #832. Frames arriving before tree start caused errors when accessing `Variables`.

### DIFF
--- a/python/pyrogue/_DataReceiver.py
+++ b/python/pyrogue/_DataReceiver.py
@@ -64,10 +64,6 @@ class DataReceiver(pr.Device,ris.Slave):
         self.ByteCount.set(0)
         super().countReset()
 
-    def _start(self):
-        super()._start()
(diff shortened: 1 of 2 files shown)


bengineerd:
@ruck314 This is ready for review. Can you check that DataReceiver works?

ruck314 (approved):

PR #853, opened 2022-02-28
ESROGUE-569 - Add function to dump entire tree including Variable attributes to dict

bengineerd:
### Description

A new `treeDict()` method has been added to `Root` to get the entire tree, including Variable attributes, as a recursive dict.
A new `properties()` method has been added to `BaseVariable` that returns a dictionary of all the `Variable`'s relevant properties.

### Details

The new method works by calling `_getDict()` underneath. A new `properties` parameter has been added to `_getDict()` to determine whether the dictionary has all the Variable properties or just a `VariableValue`.

This might not be a great way to implement this, and we should discuss it.

### JIRA

https://jira.slac.stanford.edu/browse/ESROGUE-569

### DIFF
--- a/python/pyrogue/_Command.py
+++ b/python/pyrogue/_Command.py
@@ -206,7 +206,7 @@ class BaseCommand(pr.BaseVariable):
     def _setDict(self,d,writeEach,modes,incGroups,excGroups,keys):
         pass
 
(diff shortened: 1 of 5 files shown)


valmar (approved):
I cleaned up all flake8 warning. Works as advertised and, to my limited knowledge of the project, the code looks good

PR #864, opened 2022-04-14
Update _FileReader.py

ruck314:
### Description
- Patches the code for the header size calculation with respect to the latest python collection package release


slacrherbst (approved):

bengineerd:
This should use `dataclass` instead of `namedtuple`, since the data is mutable.

bengineerd:
@ruck314 Can you verify that this still works after my changes? I don't seem to have a rogue data file to test it with.

Issue #862, opened 2022-04-07
RSSI Controller std::thread memory leak


PrimozLavric:
In `rpr::Controller::start` a thread is created using new, but never freed:
https://github.com/slaclab/rogue/blob/main/src/rogue/protocols/rssi/Controller.cpp#L146

Proposed fix:
Add `delete thread_` to `rpr::Controller::stop` or better avoid creating a thread using `new`.

bengineerd:
I created https://github.com/slaclab/rogue/pull/866 to fix this.

Thank you for these reports. We appreciate contributions from external collaborators.

bengineerd:
Resolved in #866.

PR #866, opened 2022-04-20
ESROGUE-578 - Delete thread created with new

bengineerd:
### Description

In `rpr::Controller::start` a thread was created using new, but never freed. The `stop` method now calls `delete thread_` to make sure it gets freed.

### JIRA

https://jira.slac.stanford.edu/browse/ESROGUE-578

### Related

https://github.com/slaclab/rogue/issues/862


bengineerd:
This fix seems to create a lockup that causes the test script to run forever.

PR #867, opened 2022-04-22
Add optional index param to Variable.value()

bengineerd:
### Description

You can now call `var.value(index=x)`. Not sure why this wasn't already there.


slacrherbst:
I am surprised this was not already there.

slacrherbst (approved):

bengineerd:
I was surprised too. It looked like it was there and then was commented out. Seems to work fine, so maybe just something that got overlooked.

PR #868, opened 2022-04-22
Add functions for operating on custom list of Blocks

bengineerd:
### Description

This change allows write, verify, read, and check to be performed on a custom list of Blocks.
It also provides a LinkVarable.getBlocks() method to get a list of all Blocks held by the LinkVariable dependencies.
You can then use the new Block functions to efficiently operate on this list of Blocks.

### Details

This change is particularly useful for LinkVariables that aggregate a group of RemoteVariables. It allows for those RemoteVariables to be accessed efficiently without blocking (`checkBlocks()`) between each RemoteVariable read/write/verify.

Example:
```python
class GroupLinkVariable(pr.LinkVariable):
    def __init__(self, **kwargs):
        super().__init__(
            linkedSet=self._set,
            linkedGet=self._get,
            **kwargs)
        
    def _set(self, *, value, index, write):
        # Dependencies represent each index in order
        # So just use those references to do the set accesses
        with self.parent.root.updateGroup():
            if index != -1:
                # Individual index writes happen as normal
                self.dependencies[index].set(value=value, write=write)
            else:
                for idx, (var, val) in enumerate(zip(self.dependencies, value)):
                    # Set shadow values without writing to hardware
                    var.set(value=val, write=False)

                if write is True:
                    pr.writeAndVerifyBlocks(self.depBlocks)

    def _get(self, *, index, read):
        # Dependencies represent each index in order
        # So just use those references to do the get accesses
        with self.parent.root.updateGroup():
            if index != -1:
                return self.dependencies[index].get(read=read)
            else:
                if read is True:
                    # Do a hardware read on all the dependency Blocks
                    pr.readAndCheckBlocks(self.depBlocks)
                    
                # Assign results to return variable
                ret = np.zeros(len(self.dependencies), np.float)
                for idx, var in enumerate(self.dependencies):
                    ret[idx] = var.get(read=False)

                return ret
```

### DIFF
--- a/python/pyrogue/_Block.py
+++ b/python/pyrogue/_Block.py
@@ -34,28 +34,28 @@ def writeBlocks(blocks, force=False, checkEach=False, index=-1, **kwargs):
         Allows a custom list of blocks to be efficiently written,
         similar to Device.writeBlocks(). """
     for b in blocks:
-        b._startTransaction(type=rim.Write, forceWr=force, checkEach=checkEach, index=index)
+        startTransaction(b, type=rim.Write, forceWr=force, checkEach=checkEach, index=index, **kwargs)
 
 def verifyBlocks(blocks, checkEach=False, **kwargs):
     """ Helper function for verifying a list of blocks.
         Allows a custom list of blocks to be efficiently verified without blocking
         between each read, similar to Device.verifyBlocks(). """
     for b in blocks:
-        b._startTransaction(type=rim.Verify, checkEach=checkEach)
+        startTransaction(b, type=rim.Verify, checkEach=checkEach, **kwargs)
 
 def readBlocks(blocks, checkEach=False, **kwargs):
     """ Helper function for reading a list of blocks.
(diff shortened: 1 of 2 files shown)


slacrherbst:
we may want to insist and documentation and examples for new interfaces such as this :)

slacrherbst (approved):

bengineerd:
@slacrherbst I'm not sure I like the way `LinkVariable.getBlocks()` is implemented. I think a static list of dependent blocks should be generated in `LinkVariable._finishInit()`. I'd forgotten about the existence of `_finishInit()`.

Also, do you think it would make sense to have `LinkVariable.writeBlocks()`, etc. It's sort of cleaner from an API standpoint, but also might be confusing since it wouldn't be called in the recursive `Device.writeBlocks()` call.

The whole issue here is that `LinkVariables` can have dependencies that span many parts of a tree, and it is useful to have methods that act on all of those underlying Blocks at once.

PR #872, opened 2022-05-03
bug fix for XVC.stop()

ruck314:
### Description
- [Using the detach() instead of join() for stopping the thread]( https://stackoverflow.com/questions/37015775/what-is-different-between-join-and-detach-for-multi-threading-in-c)
- Setting `thread_ = NULL;` (same method as rssi.Controller.stop())


slacrherbst:
I am not sure I agree with this.

ruck314:
Yeah, I don't think this is right approach after reading more about detach() as well.  I'll close the PR.

PR #876, opened 2022-07-22
Update _Model.py

ruck314:
#### Description
- Adding UFixed model
- Updated Fixed model's comments for "bitSize" and "binPoint"


bengineerd (approved):
I assume you've tested this yourself and the existing rim.Fixed C++ code handles the conversion appropriately.

Thanks for updating the docstrings.

ruck314:
I tried it out and seems to be doing the correct behavior.

PR #885, opened 2022-09-02
Enhancements to DebugTree GUI

bengineerd:
### Description

- Variable units are now shown right-justified in the same column as the value, rather than in their own column. This is much cleaner.
- Each Device has a "Read" command button to recursively read that Device.
- Column auto-expand works a bit better. Still could use some more work.

### Details

Here is a picture of the new GUI.
image


ruck314:
@bengineerd It crashed when I tried to use this branch on an existing software project:

bengineerd:
@ruck314 Sorry. Try again now.

ruck314:
@bengineerd Looks good now:

ruck314 (approved):

PR #886, opened 2022-09-13
Temporary fix to build with python 3.10

valmar:
Using the conda package from conda-forge avoid the "merge bug" when building rogue for python 3.10.:

https://github.com/conda/conda/issues/11442

I could build rogue for python 3.10 this way. I think this should be reverted when the bug above is closed


valmar:
Closing in favour of a pull request from a branch, rather than a fork

PR #897, opened 2022-10-20
Change pyqt4 setResizeMode() call to pyqt5 setSectionResize()

bengineerd:
### Description

Fixes a bug introduced in v5.15.2 by #895. A deprecated method (`setResizeMode()`) was called. In pyqt5, the correct method is `setSectionResizeMode()`. This was not caught because some versions of pyqt5 seem to support the deprecated `setResizeMode()` method.

### JIRA

https://jira.slac.stanford.edu/browse/ESROGUE-606

### Related

#895

### DIFF
--- a/python/pyrogue/pydm/widgets/debug_tree.py
+++ b/python/pyrogue/pydm/widgets/debug_tree.py
@@ -220,7 +220,7 @@ class DebugHolder(QTreeWidgetItem):
         self.setToolTip(0,self._var.description)
 
         self.setText(1,self._var.mode)
-        self.setText(2,self._var.typeStr)
+        self.setText(2,f'{self._var.typeStr}   ') # Pad to look nicer
         self.setToolTip(0,self._var.description)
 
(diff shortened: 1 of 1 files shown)


ruck314:
Confirmed that it works:
```bash
ruckman@rdsrv310:~/projects/snl-trans-FES/software$ python scripts/devGui.py
Rogue/pyrogue version v5.15.2-1-g00328876. https://github.com/slaclab/rogue
Start: Started zmqServer on ports 9099-9101
    To start a gui: python -m pyrogue gui --server='localhost:9099'
    To use a virtual client: client = pyrogue.interfaces.VirtualClient(addr='localhost', port=9099)
Root.PreProcessingTester.LoadThreshold()
ZmqClient::setTimeout: Setting timeout to 1000 msecs, waitRetry = 1
Connected to Root at localhost:9099
Running GUI. Close window, hit cntrl-c or send SIGTERM to 41359 to exit.
```

ruck314 (approved):

PR #904, opened 2023-01-19
Update anaconda.rst

ruck314:
### Description
- Removed `-c pydm-tag` because pydm is now in conda forge
- Updated example from rogue v5.8.0 to v5.16.0 (more current release version)

### DIFF
--- a/docs/src/installing/anaconda.rst
+++ b/docs/src/installing/anaconda.rst
@@ -55,7 +55,7 @@ Alternatively you can install a specific released version of Rogue:
 
 .. code::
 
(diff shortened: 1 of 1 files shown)


slacrherbst (approved):

ruck314:
@slacrherbst & @bengineerd: Pietro and I just went through these instruction and determine that rogue v5.16.0 has a version dependence with pydm.  Conda picks old pydm version (1.15.1) that doesn't work with rogue v5.16.0.  So the conda build has to manually enforce the newer pydm version (1.18.0)

PR #929, opened 2023-04-11
Remove Epics3 Support

slacrherbst:
This PR removes epics3 / cas support from Rogue. This also then simplifies the docker generation to no longer use the base docker that had epics pre-installed. 

This is a PR into the rogue_v6 development branch.


python/pyrogue/examples/__main__.py
```diff
@@ -24,14 +24,6 @@
...
-parser.add_argument(
-    "--epics3",
-    action   = 'store_true',
-    required = False,
-    default  = False,
-    help     = "Enable EPICS 3",
-)
-
 parser.add_argument(
     "--epics4",
```
ruck314, line 28: @slacrherbst Does this arg go away as well since only EPICS4 now?

bengineerd (approved):

PR #938, opened 2023-04-28
Fix the way nativeType and ndType are setup

slacrherbst:
In the previous code nativeType and ndType were setup the first time the nativeType() property was accessed.

This caused inefficiency as well as a potential lockup situation.

This updated code sets up nativeType and ndType in the _finishInit() routine.

This PR also changes the order to finishInit() is not called entire the entire tree is finished building.

### DIFF
--- a/python/pyrogue/_Variable.py
+++ b/python/pyrogue/_Variable.py
@@ -34,7 +34,7 @@ def VariableWait(varList, testFunction, timeout=0):
     The test function is passed a dictionary containing the current
     variableValue state index by variable path
     i.e. w = VariableWait([root.device.var1,root.device.var2],
-                          lambda varValues: varValues['root.device.var1'].value >= 10 and \
+                          lambda varValues: varValues['ooot.device.var1'].value >= 10 and \
                                             varValues['root.device.var1'].value >= 20)
 
     Parameters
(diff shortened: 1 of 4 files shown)


python/pyrogue/_Variable.py
```diff
@@ -34,7 +34,7 @@ def VariableWait(varList, testFunction, timeout=0):
     The test function is passed a dictionary containing the current
     variableValue state index by variable path
     i.e. w = VariableWait([root.device.var1,root.device.var2],
-                          lambda varValues: varValues['root.device.var1'].value >= 10 and \
+                          lambda varValues: varValues['ooot.device.var1'].value >= 10 and \
```
ruck314, line 37: @slacrherbst is `ooot` a typo?

ruck314:
@slacrherbst Is this PR intended to eventual merge with pre-release or roguev6 branch?

python/pyrogue/_Variable.py
slacrherbst: hmmm. that is odd. I did not even change that line

slacrherbst: Ok fixed. I must have fat fingered something

ruck314 (approved):

PR #954, opened 2023-06-29
Make nodeList recursive to match deviceList and variableList

slacrherbst:
Make nodeList recursive to match deviceList and variableList


ruck314 (approved):

bengineerd:
I guess if you want a non-recursive list you just do 
```python
list(self.nodes.items())
```

bengineerd (approved):

PR #959, opened 2023-07-14
saveAddressMap() removal

ruck314:
### Description
- Removed in preparation of the new rogue C++ API


bengineerd:
Is this not still useful as a debug tool?

slacrherbst:
This removes too much. We want to keep the csv address map and simply remove the header generation option. The csv address map is useful for us to look at the locations of all of the variables. I will update.

slacrherbst:
This is the alternative. I will expand this branch to also remove the other related code:
https://github.com/slaclab/rogue/compare/pre-release...rem_header

PR #972, opened 2023-09-20
Add threshold to rate test 

slacrherbst:
This PR applies a threshold to the test_rate python script. This can be used to determine if changes have significant impact on the variable access rates.

### DIFF
--- a/pip_requirements.txt
+++ b/pip_requirements.txt
@@ -19,3 +19,4 @@ gitpython
 pygithub
 pydm
 matplotlib
+hwcounter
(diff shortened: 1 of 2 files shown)


ruck314 (approved):

ruck314:
@slacrherbst Does `hwcounter` need to be included in `conda-recipe/meta.yaml` as a new dependency?

slacrherbst:
> @slacrherbst Does `hwcounter` need to be included in `conda-recipe/meta.yaml` as a new dependency?

No. It has been added to pip_requirements.txt which is used by the CI. We don't see to force it into anaconda.

PR #973, opened 2023-09-20
Add warning when loading a configuration and variable mode is not enabled

slacrherbst:
By default loading a configuration does not apply to WO variables. Instead of failing silently this PR will generate a warning message when an entry exists in the configuration file, but will not be used because the variable has a mode which is not enabled for the configuration load.

This also fixes a bug found in the display value for the DataReceiver.

### DIFF
--- a/python/pyrogue/_DataReceiver.py
+++ b/python/pyrogue/_DataReceiver.py
@@ -54,6 +54,8 @@ class DataReceiver(pr.Device,ris.Slave):
 
         self.add(pr.LocalVariable(name='Data',
                                   typeStr=typeStr,
+                                  disp='',
+                                  groups=['NoState'],
                                   value=value,
                                   hidden=hideData,
                                   description='Data Frame Container'))
--- a/python/pyrogue/_DataReceiver.py
+++ b/python/pyrogue/_DataReceiver.py
@@ -55,7 +55,7 @@ class DataReceiver(pr.Device,ris.Slave):
         self.add(pr.LocalVariable(name='Data',
                                   typeStr=typeStr,
                                   disp='',
-                                  groups=['NoState'],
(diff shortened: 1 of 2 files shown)


python/pyrogue/_DataReceiver.py
```diff
@@ -54,7 +54,6 @@ def __init__(self,
 
         self.add(pr.LocalVariable(name='Data',
                                   typeStr=typeStr,
-                                  disp='',
```
ruck314, line 57: @slacrherbst Was removing line#57 (disp='',) an intentional change for this PR?

slacrherbst: "This also fixes a bug found in the display value for the DataReceiver."

The problem was when we dumped the state the numpy array in the DataReceiver was being exported as a blank line which caused the system to choke when loading in the state as a configuration.

slacrherbst: but I just realized this is not the correct fix.

ruck314: @slacrherbst Do we want the default to be `groups=['NoState','NoServe','NoStream','NoStatus'],`?  

I am not failiar with all the groups stuff in rogue.  But for some reason you added it to the EPICS data receiver for llrf (see  URL below).  Just checking if we need it here as well.
https://github.com/slaclab/darpa-accel-llrf-phase-1p5/blob/main/firmware/python/darpa_accel_llrf_phase_1p5/_DbgProcessor.py#L32-L35

slacrherbst: I think you are right about NoStream
. NoStatus does not exist.

bengineerd (approved):

ruck314 (approved):

PR #979, opened 2023-12-06
 test_block_overlap.py update

ruck314:
### Description
- Adding corner case to test_block_overlap.py for non-4kB aligned base offset
- Fix bug found when running test_block_overlap.py

#### Overlap bug fix
Rogue was implicitly requiring that Device offsets fall on orthogonal "power-of-two" boundaries, e.g. 0x123000, 0xF00, where the LSB zeros indicate the address space of the Device. For most firmware with registers on an AXI-Lite bus, this will always be the case. But a corner case was recently encountered where an I2C peripheral requires a very odd offset. This change will support such offsets without breaking existing functionality.


slacrherbst:
I added a few more changes in this commit which address other places where the offset is orred instead of added:

https://github.com/slaclab/rogue/pull/979/commits/42ec83e4116fe5bf77e82643041466ab4a57e7b2

bengineerd (approved):

bengineerd:
Is this ready to merge?

PR #982, opened 2024-01-10
Fix nodeList method in Node and add global node count tracking

slacrherbst:
Fixes the Node method nodeList which was broken.

Also adds a class counter for Node to determine how many nodes exist in the system as a scaling metric.


bengineerd (approved):

bengineerd:
This was mistakenly pushed to main instead of pre-release, then undone with a force push.
New PR is #983

PR #985, opened 2024-02-06
Fix suggested default address. Fix broken address/port lookup

slacrherbst:
This updates the documentation to suggest using 127.0.0.1 for the zmqServer listen address. This will avoid getting annoying messages from cyber security port scanners.

We still need to properly document ZmqServer to explain the various addr arg options.

I also fixed a bug where the address lookup always returns localhost, which will break the gui when users define options other than * or 127.0.0.1


ruck314 (approved):
Confirmed that this change works and compatible with my v6.1.1 build:
https://github.com/slaclab/kek-bpm-rfsoc-dev/commit/fab69be85a5beb8b89f9f6df5bf7f8993c5ecc9d

PR #999, opened 2024-04-22
AxiStreamDma: exposing all the TX/RX diagnostic ioctl() to python

ruck314:
### Description
Also added `AxiStreamDmaMon` python class for creating the LocalVariables

```python
self.dmaStream = rogue.hardware.axi.AxiStreamDma(args.dev,(0x100*0)+0,1)                
self.add(pyrogue.hardware.axi.AxiStreamDmaMon(
    axiStreamDma = self.dmaStream,
    expand       = True,
))
```


bengineerd:
This is awesome. Is there any other driver information we might want to expose while we're at it?

ruck314:
image
@bengineerd I added more RX buffer diagnostics but it will require updating the kernel driver:
https://github.com/slaclab/aes-stream-drivers/pull/128

bengineerd (approved):

PR #1008, opened 2024-06-19
Post-build step to remove egg files

ruck314:
### Description
- Egg files from PIP install can mess up python environments.
- Make clean should remove the egg files.


slacrherbst (commented):
I think the egg clean needs to come before the setup call.

slacrherbst (approved):

PR #1015, opened 2024-07-19
Bug fix for Block.cpp

ruck314:
### Description
- bug fix for index=-1 case
- When index=-1, size_ should be tSize = (highByte - lowByte) + 1


ruck314:
out of bounds (> numValues) and -1 are treat the same and return the full array range. as calculated previously in lowTranByte and highTranByte

PR #1021, opened 2024-08-17
Fix RSSI sgement size calculation

slacrherbst:
The current segment size calculation does not include enough space for the RSSI header. This resulted in dropped packets when running with a 1500 byte MTU.


ruck314 (approved):
@slacrherbst Approve but need to fix C++ linter errors

PR #1024, opened 2024-09-26
Improve variable update performance

slacrherbst:
This PR improves the variable update performance by making the following changes:

- Do not always check for the existence of a thread specific UpdateTracker object in the tracking list.  This check requires an expensive dictionary lookup which previously occurred within a lock context. Instead the access is called as if the entry exists and if this is a new thread the resulting exception is caught and the object is then created inside the necessary lock. This change occurs in both the entry into the updateGroup context and more importantly in the _queueUpdates call.

- Take advantage of the fact that each UpdateTrack instance is per thread and remove any locking around the update of this object.

- In the check() call in the UpdateTrack first check the context counter and update period. This check is not expensive as compared to checking the variable list length, which now only occurs when the count and period tests pass.

- Change the variable listener processing to only occur when the update worker processes the list. In the previous version, every listener for a variable was added with each set or get call which was not necessary. In the new code the listeners are first added to the existing update list before the update calls are made.

### DIFF
--- a/tests/test_rate.py
+++ b/tests/test_rate.py
@@ -16,8 +16,8 @@ import rogue.interfaces.memory
 import time
 import hwcounter
 
-import cProfile, pstats, io
-from pstats import SortKey
+#import cProfile, pstats, io
+#from pstats import SortKey
 
 #rogue.Logging.setLevel(rogue.Logging.Debug)
 #import logging
@@ -108,8 +108,8 @@ class DummyTree(pr.Root):
 
 def test_rate():
 
-    pr = cProfile.Profile()
-    pr.enable()
+    #pr = cProfile.Profile()
+    #pr.enable()
 
     with DummyTree() as root:
         count = 100000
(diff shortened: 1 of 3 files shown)


tests/test_rate.py
ruck314, line 1: @slacrherbst Do you still want to keep this commented code? Alternative you could parameterize via arg to call it again if planning to need it later

ruck314:
```python
MaxCycles = { 'remoteSetRate'   : 8.0e9,
              'remoteSetNvRate' : 7.0e9,
              'remoteGetRate'   : 6.0e9,
              'localSetRate'    : 8.0e9,
              'localGetRate'    : 6.0e9,
              'linkedSetRate'   : 9.0e9,
              'linkedGetRate'   : 8.0e9 }
  ```

@slacrherbst With the improved performance, did you want to "raise the bar" on these min. rate benchmarks that are used in CI testing?

ruck314 (approved):

tests/test_rate.py
slacrherbst: I thought of parametrizing it but did not want to do it at the cost of not getting this released.

bengineerd (approved):

PR #1030, opened 2024-11-26
Add setup_rogue.fish for fish shell

JJL772:
### Description

Adds setup_rogue.fish for the [fish shell](https://github.com/fish-shell/fish-shell)


slacrherbst (commented):
I am confused. it appears the fish shell uses the same format as bash, can't they just source the bash version?

JJL772:
The syntax is different in some cases. i.e. if statements:
```sh
# Bash:
if [ -z "$LD_LIBRARY_PATH" ]; then
    ...
fi

# Fish:
if test -z "$LD_LIBRARY_PATH"
    ...
end
```

the `export` syntax is the same, I guess for backwards compatibility.

slacrherbst:
Got it. I did not look close enough.

slacrherbst (approved):

PR #1047, opened 2025-02-20
Create helper class for setting up a remote command memory bridge

slacrherbst:
This PR creates the OsCommandMemorySlave class which enables the creation of remote operating system based transactions over a memory interface. 

Examples of its usage are included here in the examples/_OsMemSlave and associated examples/_OsMemMaster classes.

Translation between the bytes transferred over the memory interface and real python variables used in the operating system classes are handled using the Model class similar to their usage in Remote Variables.


ruck314:
@slacrherbst While this new module that you created is "fresh" in your mind, would it make sense to add some amount of document (.rst file) for github.io?

python/pyrogue/examples/_OsMemSlave.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
```
ruck314, line 1: @slacrherbst Is `#!/usr/bin/env python3` required?

ruck314 (approved):

PR #1053, opened 2025-03-26
Allow the progress and steps to update every second without polling

bengineerd:
A user needed to run without polling and found it annoying that the Process progress bar no longer updates.
This will allow  the Progress and Step variables to update at one second intervals when updated with the `_incrementSteps()` and `_setSteps()` methods.


ruck314:
We will create a JIRA to use runprocess to use update groups instead of this solution

PR #1054, opened 2025-03-28
Rogue debug UI changes

JJL772:
### Description

* Added an "Offset:BitOffset" column to the debug UI displaying register offsets relative to their parent
* Added actions to the context menu to allow toggling columns, copying text for current item, and copying the path of the selected row.
* Slight cleanup to the debug_tree.py code

If the offset column causes too much visual noise, it can be disabled by default and toggled via the context menu.

### Details
image

Screenshot_20250328_132942

### DIFF
--- a/python/pyrogue/pydm/widgets/debug_tree.py
+++ b/python/pyrogue/pydm/widgets/debug_tree.py
@@ -16,6 +16,7 @@ import pyrogue
 import pyrogue.pydm.widgets
 from pyrogue.pydm.data_plugins.rogue_plugin import nodeFromAddress
 from pyrogue.pydm.widgets import PyRogueLineEdit
+from pyrogue import RemoteVariable, BaseVariable
 
 from pydm.widgets.frame import PyDMFrame
 from pydm.widgets import PyDMLabel, PyDMSpinbox, PyDMPushButton, PyDMEnumComboBox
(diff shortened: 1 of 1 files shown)


bengineerd:
Interesting stuff!
I guess it's a personal preference, but I still like the Device "Read" button in the Command column. 
The Offset:BitOffset is useful, but I think it should be formatted 
```
Type           Field
UInt4        0x04 [3:0]
```

JJL772:
> I guess it's a personal preference, but I still like the Device "Read" button in the Command column.

oops, that wasn't an intentional change! I must've missed a column number.

I'll fix that and apply your formatting suggestion, thanks for the feedback.

JJL772:
image

ruck314:
@JJL772 
1) Should the `Offset:BitOffset` be renamed to `ByteOffset:BitRange`?
2) Can we make this  `Offset:BitOffset` hidden by default with the option to make it visible later? 
    A) It's a lot of information that users (non-developers) using the software may not want to see

JJL772:
@ruck314 done

ruck314 (approved):

bengineerd (approved):

JJL772:
Resubmitting PR as a branch on Rogue, instead of from my fork.

PR #1056, opened 2025-03-31
adding instructions for installing rogue with miniforge

ruck314:
### Description
- miniforge is a light weight alternative to anaconda
image


bengineerd:
The miniforge instructions seem to be the same as the anaconda instructions except for this step
```bash
$ wget https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh
```

Maybe we just update the anaconda install page with a note about doing this to use miniforge instead?

bengineerd (approved):

PR #1058, opened 2025-03-31
ESROGUE-701 - Don't Use Spinboxes In DeviceTree GUI Widget

bengineerd:
### Description
- We never want to use spinboxes in the PyDM debug tree


ruck314:
@bengineerd This removed the spinbox. Here's the local variable I used to test with:
```python
        self.add(pr.LocalVariable(
            name        = 'MaxFmcAdcTemp',
            description = 'Alarm Set point for overheating',
            mode        = 'RW',
            value       = 100,
            units       = 'degC',
            minimum     = -100,
            maximum     = 200,
        ))
```
However, the max/min range is not working anymore.  I was able to set a `999` value into this variable and the rogue didn't have a error and allowed the variable to get set to `999` value
image

ruck314:
Note: Local block is not doing range checking right now.  Something that we need to fix

ruck314 (approved):

PR #1060, opened 2025-04-01
Disable verify if overlapping variable is RO, WO or has verify disabled

slacrherbst:
Disable verify if overlapping variable is RO, WO or has verify disabled.


src/rogue/interfaces/memory/Block.cpp
```diff
@@ -500,6 +504,11 @@ void rim::Block::addVariables(std::vector<rim::VariablePtr> variables) {
                     setBits(verifyMask_, (*vit)->bitOffset_[x], (*vit)->bitSize_[x]);
                 }
 
+                // update verify allow
+                if ((*vit)->mode_ == "RO" || (*vit)->mode_ == "WO" || ! (*vit)->verifyEn_) {
```
ruck314, line 508: @slacrherbst Should this "WO" be "RW" instead? "WO" doesn't require verifying

src/rogue/interfaces/memory/Block.cpp
```diff
@@ -540,6 +549,11 @@ void rim::Block::addVariables(std::vector<rim::VariablePtr> variables) {
                     setBits(verifyMask_, x * (*vit)->valueStride_ + (*vit)->bitOffset_[0], (*vit)->valueBits_);
                 }
 
+                // update verify allow
```
ruck314, line 552: @slacrherbst Should this "WO" be "RW" instead? "WO" doesn't require verifying

ruck314:
@slacrherbst Are you planning to update the testing CI script to exercise this new feature?

src/rogue/interfaces/memory/Block.cpp
```diff
@@ -540,6 +549,11 @@ void rim::Block::addVariables(std::vector<rim::VariablePtr> variables) {
                     setBits(verifyMask_, x * (*vit)->valueStride_ + (*vit)->bitOffset_[0], (*vit)->valueBits_);
                 }
 
+                // update verify block
+                if ((*vit)->mode_ == "RO" || (*vit)->mode_ == "WO" || !(*vit)->verifyEn_) {
```
ruck314, line 553: @slacrherbst Should this "WO" be "RW" instead? "WO" doesn't require verifying

slacrherbst:
> @slacrherbst Are you planning to update the testing CI script to exercise this new feature?

Our emulator does not have the ability to emulate RO and WO values.

PR #1064, opened 2025-04-07
Update operation of local variables

slacrherbst:
This PR addresses a number of issues related to the operation of local variables.

First, it adds min/max enforcement on local variables in the underlying LocalBlock class.

Second, it generates an error if a user tries to set min and max values on a LinkVariable.

Third is reducing the layer of calls that occur during each set and get operation. Previously, a number of unnecessary calls we made through the various Device and Block layers. This PR reduces the number of calls while still ensuring the variable updates are queued to the update worker thread in Root.

Finally, an updateGroup() wrapper is added to the RunControl and Process classes to allow a 1.0 second update of status variables with Polling disables. 

### JIRA
https://jira.slac.stanford.edu/browse/ESROGUE-704


python/pyrogue/_RunControl.py
```diff
@@ -140,10 +140,12 @@ def _run(self):
...
-        while (self.runState.valueDisp() == 'Running'):
-            time.sleep(1.0 / float(self.runRate.value()))
-            if self._cmd is not None:
-                self._cmd()
+        with self.root.updateGroup(period=1.0):
 
-            with self.runCount.lock:
-                self.runCount.set(self.runCount.value() + 1,write=False)
+            while (self.runState.valueDisp() == 'Running'):
+                time.sleep(1.0 / float(self.runRate.value()))
```
ruck314, line 146: @slacrherbst Should `time.sleep(1.0 / float(self.runRate.value()))` be `time.sleep(period / float(self.runRate.value()))` (use period arg instead of hard coded 1.0)?

slacrherbst: No run rate is in hz, so the sleep period is 1/rate.

ruck314 (approved):

PR #1066, opened 2025-04-08
Don't allow default set to RO register to result in stale state

slacrherbst:
This PR blocks a set to a RO variable from impacting the stale state of the variable or the underlying block.

This is necessary to allow a default value in a RO register.

### DIFF
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -94,7 +94,7 @@ class MemDev(pr.Device):
                     offset       = 0x000+4*i,
                     bitSize      = 8,
                     bitOffset    = 8*j,
-                    mode         = modeConfig,
+                    mode         = "RW",
                     value        = value
                 ))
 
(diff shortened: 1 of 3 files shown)


tests/test_memory.py
```diff
@@ -137,12 +137,10 @@ def __init__(self):
         self.addInterface(mc)
 
         # Add Device
-        modeConfig = ['RW','RW','RO','RO']
         for i in range(4):
             self.add(MemDev(
                 name       = f'MemDev[{i}]',
                 offset     = i*0x10000,
-                modeConfig = modeConfig[i],
```
ruck314, line 145: @slacrherbst You removed modeConfig, so now the test only covers "RW" registers instead of both "RW" and "RO". Shouldn't we still be testing the "RO" case in this script? If not, then wouldn’t it make more sense to remove the loop over 4 entirely and just instantiate MevDev once?

slacrherbst: It's not clear to me there is a clan way to test the RO registers with the limitations of the emulator. The best test relied on a bug in the code which allows RO registers to be written.

ruck314 (approved):

PR #1116, opened 2026-02-24
Enhance PyRogue docs and typing across utilities and PyDM

bengineerd:
### Description
- Add and reorganize docs for file I/O format and HLS register parser utilities:
  - add canonical Rogue file format page and cross-links from reader/writer docs
  - add HLS utility docs for `pyrogue.utilities.hls._RegInterfParser` (workflow, CLI usage, and function behavior)
- Improve typing and docstrings in utilities modules (`cpsw`, `fileio`, `prbs`, HLS parser), including clearer parameter/return docs and callback signatures.
- Perform broad PyDM/UI docstring and type hint cleanup:
  - strengthen type hints (including `QWidget`/`PyDMChannel` where applicable)
  - add/complete class and method docstrings
  - ensure class docstrings document `__init__` parameters consistently.

### DIFF
--- a/docs/src/utilities/fileio/classes/reader.rst
+++ b/docs/src/utilities/fileio/classes/reader.rst
@@ -6,6 +6,10 @@ StreamReader
 
 The StreamReader class is the utilitiy for reading data from a Rogue file in a streaming fashion.
 
+The file format consumed by StreamReader is documented here:
+
+- :ref:`utilities_fileio_format`
+
 StreamReader objects in C++ are referenced by the following shared pointer typedef:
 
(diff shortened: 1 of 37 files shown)


ruck314:
@bengineerd I wonder if we can get rid of the `python/pyrogue/utilities/hls/_RegInterfParser.py` because I don't think anyone is using it.

bengineerd:
Let's leave it for now. But I do think we either want to get rid of it or enhance it to be more useful. It is a neat idea.

ruck314 (approved):

PR #1142, opened 2026-03-18
Fix RunControl, Process, and UART worker shutdown bugs

bengineerd:
## Summary

This PR fixes several shutdown and progress-handling bugs across PyRogue's `RunControl`, `Process`, and UART memory worker, and includes follow-up fixes from PR review.

## Bug Fixes

1. **RunControl worker shutdown and restart race**

- The stop path no longer attempts to `join()` the current worker thread from inside that same thread.
- `RunControl` now keeps ownership of `self._thread` until `_run()` actually exits.
- The start path only launches a new worker when there is no live run thread.
- This closes the race where a quick stop/start transition from inside the worker could create a second concurrent `_run()` loop.

2. **UART worker queue completion on all paths**

- The UART worker now calls `self._workerQueue.task_done()` in a `finally` block for every dequeued item.
- Unexpected exceptions from `_doWrite()` or `_doRead()` are logged and converted into `transaction.error(...)`.
- This prevents `_stop()` and `Queue.join()` from hanging again if transaction handling raises.

3. **`Process.__call__(arg=...)` argument handling**

- `Process.__call__()` now writes the configured argument variable directly with `self._argVar.setDisp(arg)`.
- This fixes the `KeyError` path caused by looking up the variable object in `self.nodes`.

4. **Process progress clamping without private block mutation**

- `Progress` still clamps direct writes into the valid `0.0..1.0` range.
- The implementation no longer mutates `Progress._block._minimum` or `Progress._block._maximum`.
- Instead, `Process` now uses a private `LocalVariable` subclass for `Progress` that clamps values before the base `LocalVariable.set()` range validation runs.
- This preserves the declared `minimum` and `maximum` metadata while avoiding unsupported block-internal mutation.

5. **Step-based progress recomputation**

- Step-based progress updates still recompute `Progress` automatically when `Step` or `TotalSteps` changes directly.
- Out-of-range step counts and non-positive totals are clamped to a valid `0.0..1.0` progress result.


ruck314 (approved):

ruck314:
@bengineerd Review comments from meeting:
- Concerns that this is over optmizing a "example" class that users would us it
  - Remove these optmize
- probably _argVar and Ryan thinks that this is a bug
- probably want to keep the tread checking in _RunControl.py

python/pyrogue/_Process.py
```diff
@@ -83,8 +86,11 @@ def __init__(
             disp = '{:1.2f}',
             minimum=0.0,
             maximum=1.0,
+            localSet=self._clampProgressValue,
             pollInterval=1.0,
             description='Percent complete: 0 - 100 %.'))
+        self.Progress._block._minimum = None
```
bengineerd: This feels like a hack. I need to check on it.

PR #1146, opened 2026-03-18
Fix SqlLogger syslog decoding for VariableValue payloads

bengineerd:
## Bug Fixes

1. **SqlLogger syslog handling decoded `VariableValue`, not the underlying JSON string**

   What was broken:
   - Syslog updates from `SystemLogLast` arrived as `VariableValue`, but the logger passed that wrapper to `json.loads()`.

   Inputs that triggered it:
   - Any `SystemLogLast` update routed into SQL logging.

   Result:
   - Valid syslog entries failed to decode and log correctly.

   Fix:
   - Unwrapped the stored string payload before JSON decoding.

### DIFF
--- a/python/pyrogue/interfaces/_SqlLogging.py
+++ b/python/pyrogue/interfaces/_SqlLogging.py
@@ -113,7 +113,10 @@ class SqlLogger(object):
 
         # Syslog
         if entry[0] == self._sysLogPath:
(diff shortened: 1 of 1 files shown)


ruck314 (approved):

ruck314:
@bengineerd Should be entry.value (instead of entry[])

PR #1150, opened 2026-03-19
Add LocalVariable typeCheck flag and disable it for DataReceiver.Data

bengineerd:
## Summary
- add a `typeCheck` parameter to `LocalVariable`, defaulting to `True`
- gate the LocalBlock type-mismatch warning on that flag
- disable type checking for `DataReceiver.Data` and document the new behavior
- add tests covering the default warning path and the opt-out path
- add documentation of this feature


ruck314 (approved):
@bengineerd I confirmed that this fixed the issue that I was observing in Wave8

slacrherbst:
Given now that we can change the mode of the variable, should an incorrect type throw an exception?

PR #1152, opened 2026-03-19
Fix PyDM debug_tree name corruption after long ZMQ transactions

bengineerd:
## Summary
This PR fixes a long-standing PyDM `debug_tree` display bug that shows fully qualified Rogue channel paths in every tree row after a long-running device read completes through `ZmqServer` / `VirtualClient`.

It also includes transport-side fixes that came out of debugging that behavior:

- `VirtualClient` link-state handling now treats long in-flight requests as busy work instead of transient disconnects.
- `VirtualClient` startup now tolerates a slow initial `__ROOT__` reply instead of caching a failed half-initialized client.
- `ZmqClient` retry logging is throttled so long transactions do not emit a timeout warning every second.
- `ZmqClient` shutdown is more reliable when a client bootstrap attempt fails.

### Symptom
When a PyDM GUI is connected to a Rogue tree through `ZmqServer`, a long-running recursive read such as `ReadDevice` can leave the `debug_tree` widget in a broken state:

- the GUI shows repeated messages like `Timeout waiting for response after XX Seconds, server may be busy!`
- once the long transaction finally completes, each `debug_tree` leaf shows the full Rogue channel path instead of the short node name
- reading an individual variable from the GUI fixes that one row because it forces a fresh value update

### Root Cause
The bug was caused by an interaction between three pieces of behavior:

1. `debug_tree` renders node names with `PyDMLabel` widgets connected to `<path>/name`.
2. `PyDMLabel` falls back to displaying the raw channel string while the channel is disconnected.
3. `VirtualClient` only used publish/update traffic as proof of life. During a long blocking request, the client could go more than 10 seconds without receiving publish traffic, so it marked the link down even though the request was still legitimately in flight.

That caused the `PyDMLabel` widgets in the tree to switch from short names to the full channel string.

When the server eventually responded, the link state recovered, but static channels such as `/name` did not necessarily receive a fresh monitor update. As a result, the labels stayed stuck on the full path until some other action refreshed that specific node.

### Fix
The Rogue PyDM data plugin now explicitly re-emits the current cached value when the shared `VirtualClient` link comes back up.

That reconnect refresh restores the intended text for:

- `/name`
- `/path`
- device-name rows
- cached variable displays

This makes the `debug_tree` recover immediately once the long request finishes instead of waiting for another manual read.

### Problem
The label bug exposed a more general issue in `VirtualClient`: long-running requests were treated like disconnects because lack of publish updates was interpreted as lack of connectivity.

That is not correct for firmware co-simulation or any backend where a legitimate memory transaction can remain in flight for many seconds or minutes.

### Fix
`VirtualClient` now tracks outstanding request/reply activity separately from publish traffic.

- while a request is in flight, the client remains linked instead of declaring the server disconnected after the idle heartbeat window
- a successful request reply now counts as proof of life and refreshes the link state immediately

The client also now supports explicit timeout configuration:

- `linkTimeout`: idle timeout before the link is considered down
- `requestStallTimeout`: optional maximum age for an in-flight request before it is considered stalled

In practice:

- `linkTimeout` is the primary operational knob and is the one most applications should tune
- `requestStallTimeout` is a niche policy knob and is usually best left disabled unless a deployment has a clear upper bound for legitimate request duration

By default:

- `linkTimeout` remains 10 seconds
- `requestStallTimeout` is disabled so long co-sim transactions are treated as busy rather than dead

These settings can now be passed directly through `pyrogue.pydm.runPyDM(...)`, which preconfigures the cached `VirtualClient` instances used by the PyDM plugin.

### Problem
While testing the timeout changes, this branch also exposed an older startup bug in `VirtualClient` bootstrap.

The initial `__ROOT__` request was treated as a one-shot 1 second operation. If the server took slightly longer to answer, `VirtualClient` returned early but had already inserted itself into the shared client cache. That left a poisoned cached object with no root tree attached, and every later `VirtualClient(addr, port)` in that process reused the broken instance.

This was especially visible in PyDM because the branch now eagerly preconfigures cached `VirtualClient` instances before widgets attach.

### Fix
`VirtualClient` startup now uses a bounded bootstrap window for the initial `__ROOT__` handshake instead of a single 1 second failure path.

- the initial connect retries until a short deadline expires
- the default bootstrap timeout is 5 seconds
- the timeout can be overridden with `ROGUE_VIRTUAL_CONNECT_TIMEOUT`
- failed bootstrap clients are removed from the cache and their transport is shut down cleanly

This preserves normal startup behavior for a healthy server, avoids poisoning the shared cache, and makes PyDM attach more robust when the server is busy during startup.

### Problem
The previous retry behavior emitted a timeout log message once per receive-timeout interval, which meant an intentional long transaction produced log spam every second.

The same code path also made cleanup of a failed startup client awkward because the subscriber thread could block indefinitely waiting on traffic.

### Fix
`ZmqClient` now logs retry waits at a much lower rate:

- one warning when the client first enters the wait-retry path
- periodic "still waiting" warnings every 30 seconds
- one recovery message when the response finally arrives

It also now applies a receive timeout on the subscriber socket so failed bootstrap clients can stop cleanly during teardown.

This preserves visibility into a busy server without flooding the log during long operations and makes failed startup cleanup deterministic.

## Documentation
The user-facing docs and docstrings were updated to explain:

- why `linkTimeout` is the normal timeout to tune
- why `requestStallTimeout` is optional and often unnecessary
- how to pass these settings through `VirtualClient(...)` and `pyrogue.pydm.runPyDM(...)`
- how `ZmqClient` retry logging now behaves during long waits
- how `ROGUE_VIRTUAL_CONNECT_TIMEOUT` affects initial bootstrap only

## Tests
- `conda run -n rogue_build python -m pytest -q tests/test_virtual_client_link.py`
- `conda run -n rogue_build python -m pytest -q tests/test_virtual_client_connect.py`
- `conda run -n rogue_build python -m pytest -q tests/test_pydm_run_config.py`

`tests/test_pydm_rogue_plugin.py` was added for the PyDM reconnect refresh logic, but it requires a Qt/PyDM environment that is not available in the current shell configuration.

### DIFF
--- a/docs/src/pydm/starting_gui.rst
+++ b/docs/src/pydm/starting_gui.rst
@@ -140,6 +140,8 @@ to :py:func:`pyrogue.pydm.runPyDM`:
            title='My System',
            sizeX=1000,
            sizeY=500,
+           linkTimeout=600.0,
+           requestStallTimeout=None,
        )
 
 In practice, ``runPyDM`` always connects through the Rogue client interface
@@ -160,11 +162,29 @@ The main :py:func:`pyrogue.pydm.runPyDM` options are:
 - ``title``: Window title.
 - ``sizeX`` and ``sizeY``: Initial window size.
 - ``maxListExpand`` and ``maxListSize``: Limits used by the debug-tree view.
+- ``linkTimeout``: Idle timeout for the cached ``VirtualClient`` instances
+  used by the GUI. This is the primary knob for tolerating long busy periods.
+- ``requestStallTimeout``: Optional per-request stall policy for the cached
+  ``VirtualClient`` instances. This is disabled by default and usually only
+  makes sense when the application has a strict upper bound for valid request
+  duration.
 
 The command-line launcher defaults to ``localhost:9099``. The Python helper's
 function signature currently defaults to ``localhost:9090``, so in practice it
 is better to pass ``serverList`` explicitly rather than rely on the helper
 default.
+
+For simulation or co-simulation applications, the common pattern is to leave
+``requestStallTimeout`` disabled and increase ``linkTimeout`` to match the
+expected busy period:
+
+.. code-block:: python
+
+   pyrogue.pydm.runPyDM(
+       serverList=root.zmqServer.address,
+       linkTimeout=600.0,
+       requestStallTimeout=None,
+   )
 
 Rogue Channel Prefix
 ====================
--- a/docs/src/pyrogue_tree/client_interfaces/virtual.rst
+++ b/docs/src/pyrogue_tree/client_interfaces/virtual.rst
@@ -93,6 +93,44 @@ Link-State Monitoring
        print(client.linked)
        client.remLinkMonitor(link_monitor)
 
+Timeout Tuning
+==============
+
+``VirtualClient`` has two separate timeout concepts:
+
+- ``linkTimeout``: how long the client may go without publish updates or a
+  successful request reply before it marks the link down
+- ``requestStallTimeout``: an optional policy timeout for declaring a single
+  in-flight request stalled
+
+In practice, ``linkTimeout`` is the normal tuning knob. It is useful when a
+system may legitimately spend a long time servicing a request, for example
+when a Rogue tree is backed by firmware co-simulation.
+
+``requestStallTimeout`` is disabled by default and is often best left that way.
+It is only useful when the deployment has a known upper bound for legitimate
+request duration and wants the client to declare a request hung after that
+threshold.
+
+.. code-block:: python
+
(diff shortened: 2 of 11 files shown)


ruck314:
@bengineerd Here's the JIRA ticket related to this bug: https://jira.slac.stanford.edu/browse/ESROGUE-620

slacrherbst:
Other than my once comments. this appears ok

ruck314:
@bengineerd Need to fix broken CI

bengineerd:
On it

bengineerd:
I have also tested this on one of my project. It seems to be working.

ruck314 (approved):

PR #1155, opened 2026-03-23
Fix RSSI maxPayload sizing

bengineerd:
## Summary
- restore RSSI segment sizing to use the full UDP payload budget via `udp.maxPayload()`
- update `UdpRssiPack`, the UDP packetizer regression test, and the RSSI/UDP docs to match the controller semantics
- clarify in the RSSI client/server headers that `segSize` already includes the RSSI header budget on the lower transport

## Root Cause
RSSI `segSize` is already the total on-wire segment budget. Subtracting 8 at the call site double-counted the RSSI header and reduced the effective segment size, which broke the `CoreV2` non-jumbo UDP packetizer path under reordering.

## Verification
- ran `tests.test_udpPacketizer.test_data_path()` in the branch worktree
- verified all four combinations pass, including `ver=2 jumbo=False`


bengineerd:
This was a red herring.

PR #1161, opened 2026-03-29
Add epicsV7 protocol using pythonSoftIOC (softioc)

ruck314:
## Summary

- Adds `python/pyrogue/protocols/epicsV7.py` — a new EPICS protocol module using [pythonSoftIOC](https://diamondlightsource.github.io/pythonSoftIOC/master/index.html) (`softioc`) in place of `p4p`
- `EpicsPvServer` API is identical to `epicsV4.py` (same constructor, `list()`, `dump()`) to ease migration
- Adds `tests/test_epicsV7.py` with full type-coverage tests mirroring `test_epics.py`
- Updates pip/conda/conda-recipe deps to include `softioc`; `p4p` is retained as the PVA test client
- Adds RST documentation under `docs/src/built_in_modules/protocols/epicsV7/`

## Changes

- `python/pyrogue/protocols/epicsV7.py` — new softioc-based EPICS server
- `tests/test_epicsV7.py` — full test suite mirroring `test_epics.py`; commands tested via `ctxt.put()` (not `ctxt.rpc()`)
- `pip_requirements.txt`, `conda.yml`, `conda-recipe/meta.yaml` — add `softioc` dependency
- `docs/src/built_in_modules/protocols/epicsV7/` — `index.rst`, `epicspvserver.rst`, `epicspvholder.rst`

## Design notes

- `epicsV4.py` is unchanged (backward compatibility)
- IOC runs in a daemon background thread via `asyncio_dispatcher.AsyncioDispatcher`; `iocInit()` is called at most once per process
- Commands are exposed as `longOut` records; `caput value=0` → no-arg call, `caput value=N` → `command(N)`
- Neither softioc nor p4p exposes a Python callback on client `caget`/`ctxt.get()`. Both V4 and V7 serve the most recently pushed value. Live hardware values require PyRogue auto-polling (`pollInterval` on variables or `pollEn=True` on Root) or explicit device reads
- Graceful `ImportError` warning if `softioc` is not installed

## Test plan

- [x] `pytest tests/test_epicsV7.py` in an environment with `softioc` and `pyrogue` installed
- [x] Confirm existing `pytest tests/test_epics.py` still passes (epicsV4 unchanged)


python/pyrogue/protocols/epicsV7.py
```diff
@@ -0,0 +1,411 @@
...
+            else:
+                self._record = builder.longIn(self._suffix)
+            return
+
+        # --- UInt64 / Int64 → float (TYPE-02) ---
+        if typeStr in ('UInt64', 'Int64'):
+            if is_writable:
+                self._record = builder.aOut(self._suffix, on_update=self._on_put)
+            else:
+                self._record = builder.aIn(self._suffix)
```
JJL772: You can use `int64In` and `int64Out` here I believe. Seems like `softioc` supports it: https://diamondlightsource.github.io/pythonSoftIOC/master/reference/api.html?highlight=int64in#softioc.builder.int64In

python/pyrogue/protocols/epicsV7.py
```diff
@@ -0,0 +1,411 @@
...
+                self._record = builder.mbbIn(self._suffix, *enum_strings)
+            return
+
+        # --- string / list / dict / None (TYPE-07) ---
+        if (v.nativeType is list or v.nativeType is dict or
+                typeStr in ('str', 'list', 'dict', 'NoneType') or typeStr == ''):
+            if is_writable:
+                self._record = builder.stringOut(self._suffix, on_update=self._on_put)
+            else:
+                self._record = builder.stringIn(self._suffix)
```
JJL772: It might be worth using lsi/lso (long string in/out) here, because stringin/stringout have a very 1980s length limit of 40 chars. softioc appears to support lso/lso with longStringIn/longStringOut API: https://diamondlightsource.github.io/pythonSoftIOC/master/reference/api.html?highlight=int64in#softioc.builder.longStringIn

python/pyrogue/protocols/epicsV7.py
```diff
@@ -0,0 +1,802 @@
...
+        enum_strings = list(var.enum.values())
+        if len(enum_strings) <= 16:
+            return 'enum'
+        else:
+            return 's'  # large-enum fallback: matches longStringOut in softioc
+    if (var.nativeType is list or var.nativeType is dict or
+            typeStr in ('str', 'list', 'dict', 'NoneType') or typeStr == ''):
+        return 's'
+    if typeStr == 'Bool':
+        return 'i'
```
JJL772: Use `?` (bool), unless you explicitly want to represent `bool` as int

ref: https://epics-base.github.io/p4p/values.html#type-definitions

python/pyrogue/protocols/epicsV7.py
```diff
@@ -0,0 +1,802 @@
...
+    if typeStr in ('UInt64', 'Int64'):
+        return 'l'
```
JJL772, line 169: For `UInt64` use `L` (u64) to avoid data loss

python/pyrogue/protocols/epicsV7.py
```diff
@@ -0,0 +1,802 @@
...
+    if (typeStr == 'int' or
+            any(typeStr.startswith(p) for p in
+                ('UInt8', 'UInt16', 'UInt32', 'Int8', 'Int16', 'Int32'))):
+        return 'i'
```
JJL772, line 177: Use `I` (u32) for `UInt32` to prevent data loss.

The other types can be specialized as well, but you can also leave them as normal int:
* int8 -> b
* uint8 -> B
* int16 -> h
* uint16 -> H

https://epics-base.github.io/p4p/values.html#type-definitions

python/pyrogue/protocols/epicsV7.py
```diff
@@ -0,0 +1,802 @@
...
+        nt = p4p.nt.NTEnum(display=False, control=False, valueAlarm=False)
+        enum_strings = list(var.enum.values())
+        disp = varVal.valueDisp if varVal.valueDisp is not None else ''
+        try:
+            idx = enum_strings.index(disp)
+        except ValueError:
+            idx = 0
+        iv = {'choices': enum_strings, 'index': idx}
+    elif pva_type == 's':
+        nt = p4p.nt.NTScalar('s', display=False, control=False, valueAlarm=False)
```
JJL772: Any reason to avoid `valueAlarm`? It looks like alarm severities are handled by the `softioc` portion of this code.

ruck314: valueAlarm=False was originally set to keep the PVA structure minimal, but there's no real reason to suppress it. 

In fc4358c, removed the alarm fields are now included in the NTScalar structure. Alarm propagation to the PVA side is not wired up yet (the softioc record handles severity on the CA side), but leaving the fields present is correct and allows it to be added later without a structure change.

python/pyrogue/protocols/epicsV7.py
ruck314: Add these types in fc4358c

python/pyrogue/protocols/epicsV7.py
ruck314: Add these types in fc4358c

python/pyrogue/protocols/epicsV7.py
JJL772: if you need an example of how to work with the alarm metadata, I have some code here: https://github.com/lume-science/lume-pva/blob/main/lume_pva/variables.py#L89-L108

The display meta (i.e. `display.units`) could also be handy if Rogue variables have such a field attached

JJL772 (approved):
LGTM

python/pyrogue/protocols/epicsV7.py
ruck314: Addressed in 9b90be3cac083ec36c72675ef2a3912243ca4f9d

PR #1197, opened 2026-05-04
feat(pyrogue::Device): move YAML config loading from Root to Device

ruck314:
## Summary

- Move `saveYaml`, `loadYaml`, `setYaml` from `Root` to `Device` so any device in the tree can load/save YAML configuration using device-relative paths
- Add `_applyYamlDict()` dispatch hook: Device resolves relative child names via `nodeMatch()`; Root overrides to resolve absolute dotted paths via `_setDictRoot()` for backward compatibility
- Add `_writeConfig()`: Device scopes write/verify/check to its subtree; Root overrides to delegate to full-tree `_write()`
- Root overrides `loadYaml`/`setYaml` to add `InitAfterConfig` behavior
- All existing Root-level YAML operations, `SaveConfig`/`LoadConfig` commands, and YAML file formats work unchanged

## Device-level YAML format

```yaml
# Load on a specific device — uses device name or child names as keys
MyDevice:
  Variable1: 10
  SubDevice:
    Variable2: 20

# Or target children directly
Channel[*]:
  Gain: 100
```

## Test plan

- [x] 22 new tests in `tests/core/test_device_yaml_configuration.py` covering device-level save/load/set, array wildcards, scoped writes, ForceWrite, round-trips, zip/directory support, error handling
- [x] All 252 existing core tests pass unchanged (backward compat verified)
- [x] Pre-existing `test_pydm_rogue_plugin.py` failure is unrelated (confirmed on main)
- [x] CI pipeline passes

### DIFF
--- a/docs/src/pyrogue_tree/core/block_operations.rst
+++ b/docs/src/pyrogue_tree/core/block_operations.rst
@@ -161,10 +161,12 @@ How YAML Configuration Uses These Methods
 YAML configuration loading is one of the main places where users encounter
 bulk block operations indirectly.
 
-For ``Root.loadYaml(..., writeEach=False)`` and ``Root.setYaml(..., writeEach=False)``,
+For ``Device.loadYaml(..., writeEach=False)`` and ``Device.setYaml(..., writeEach=False)``,
 PyRogue first stages values into Variables with ``setDisp(..., write=False)``.
-Only after that full YAML payload has been applied to the tree does Root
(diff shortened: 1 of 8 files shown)


bengineerd:
I updated the regression tests. The existing Root yaml tests were duplicating a lot of the new Device yaml tests.
I also expanded coverage for `incGroups`/`excGroups`.

bengineerd (approved):

PR #1229, opened 2026-05-07
feat(pyrogue): add pre-write event listeners (ESROGUE-743)

ruck314:
## Summary

- Adds synchronous pre-write listeners that fire before hardware writes, enabling callers to block writes (via `WriteBlockedError`) or modify the value being written
- Supports both variable-level (`var.addPreWriteListener(...)`) and device-level (`dev.addPreWriteListener(...)`) registration
- Device-level listeners fire before variable-level; all must pass (AND logic); value modifications chain in registration order
- Listeners receive a configurable state snapshot of other variables for decision-making

## Motivation

Vision Research CXP-over-Fiber cameras enter an indeterminate state (requiring power cycle) if register writes occur during image acquisition. The existing post-write listener infrastructure cannot prevent such writes. Pre-write listeners provide a clean mechanism to guard hardware from unsafe writes based on system state.

## API

```python
import pyrogue as pr

def guard(path, value, state):
    if state.get(acq_var.path) is True:
        raise pr.WriteBlockedError(path, "camera acquiring")
    return None  # or return modified value

var.addPreWriteListener(guard, stateVars=[acq_var])
dev.addPreWriteListener(guard, stateVars=[acq_var])  # device-wide
```

## Files Changed

- `python/pyrogue/_Variable.py` — `WriteBlockedError`, storage, registration, dispatch, injection into `set()`/`post()`
- `python/pyrogue/_Device.py` — device-level storage and registration
- `tests/core/test_pre_write_listeners.py` — 17 new test cases
- `docs/src/pyrogue_tree/core/variable.rst` — narrative docs
- `docs/src/pyrogue_tree/core/device.rst` — narrative docs

## Test Plan

- [x] 17 new pytest cases covering: allow, block, value modification, chaining, state snapshots, device-level, ordering, removal, `write=False` bypass, `post()`, error propagation, RemoteVariable paths
- [x] Full core test suite passes (268 tests, 0 failures)
- [x] flake8 clean
- [x] CI full build + test (Linux, macOS)


scompa18:
@ruck314 I have tested this on a working setup and it seems to be working. Writes are prevented based on a state, as one would expect

scompa18 (approved):
Tested on running setup and seems to be working

PR #1233, opened 2026-05-11
Fix RemoteCommand, RemoteVariable.post, and Device block ops to wait for transaction completion

ruck314:
### Description

This PR started as a fix for a transaction-wait forwarding bug: several PyRogue call sites passed `checkEach=` into `pr.startTransaction()`, where it was silently absorbed by `**kwargs` instead of causing the underlying memory transaction to wait for completion. That allowed back-to-back operations, especially `BaseCommand.toggle()` (`set(1); set(0)`), to race the shared block buffer before asynchronous memory slaves had consumed the first transaction.

While fixing that path, the transaction terminology has been cleaned up so the Python-facing API now uses `wait` for transaction completion. The older `check` names were really describing "wait for the transaction and surface errors," so they are kept as deprecated compatibility aliases rather than removed.

### API Updates

Preferred names now include:

- `pr.startTransaction(..., wait=...)`
- `pr.waitTransaction(...)`
- `pr.waitBlocks(...)`
- `pr.readAndWaitBlocks(...)`
- `waitEach=` for block helper and `Device` block operations
- `Device.forceWaitEach`
- `Variable.set/get/write(..., wait=...)`

Deprecated compatibility aliases remain supported and emit `DeprecationWarning`:

- `check=` -> `wait=`
- `checkEach=` -> `waitEach=`
- `checkTransaction()` -> `waitTransaction()`
- `checkBlocks()` -> `waitBlocks()`
- `readAndCheckBlocks()` -> `readAndWaitBlocks()`
- `forceCheckEach` -> `forceWaitEach`

Supplying conflicting values such as `wait=True, check=False` raises `ValueError`.

`LinkVariable` callbacks may accept either `wait` or `check`; PyRogue forwards both names so new callback code can use `wait` while existing callback code remains compatible.

### Implementation Notes

- Internal PyRogue call sites now forward `wait=` / `waitEach=` correctly into transaction helpers.
- The original RemoteCommand/RemoteVariable/Device race is covered by regression tests that assert the correct wait flag reaches `pr.startTransaction()`.
- Docs now describe block operations as initiating memory transactions and waiting for transaction completion.
- C++ `Block` comments were clarified: `checkTransaction()` both waits for pending completion and checks transaction/verify results. The C++ public names are unchanged.

### Validation

- Focused Python core subset: `40 passed`
- C++ tests: `ctest --test-dir build -L cpp --output-on-failure` -> `14/14 passed`
- `git diff --check` clean


bengineerd:
Plan to change `check` to `wait`.

slacrherbst (approved):

PR #1247, opened 2026-05-29
feat: make blocking/non-blocking a call-site argument on Command

tristpinsm:
## Summary

Replaces the construction-time `nonBlocking=True` flag from PR #1244 with a call-site `blocking=False` keyword argument on `__call__`/`call`/`callDisp`. This addresses the design concern raised in [#1244 review](https://github.com/slaclab/rogue/pull/1244#discussion): whether to block is a function of the calling context, not a property of the command itself.

Key changes:

- **`blocking` kwarg on `__call__`/`call`/`callDisp`** (default `True`) — any Command can now be called non-blocking without construction-time opt-in
- **`@pr.expose` properties** (`running`, `result`, `error`) on `BaseCommand` — status is accessed directly on the command node instead of via injected sibling `LocalVariable`s, keeping the rogue tree clean
- **`SimpleClient.exec`** gains a `blocking` kwarg, forwarded through ZMQ
- **No sibling variable injection** — removes `_rootAttached` override, the `list()` snapshot workaround in `Device._rootAttached`, and all `_nonBlocking`/`_varRunning`/`_varResult`/`_varError` internals
- **`RemoteCommand`** no longer accepts `nonBlocking` — the `blocking` kwarg works uniformly through `__call__`

### Usage

```python
# Server — no construction-time flag needed
self.add(pr.LocalCommand(name='ApplyConfig', function=my_fn, retValue=''))

# Client — caller decides whether to block
cmd = client.root.Dev.ApplyConfig
cmd(blocking=False)          # returns None immediately

while cmd.running:           # exposed property
    time.sleep(0.1)

if cmd.error:                # exposed property
    raise RuntimeError(cmd.error)
print(cmd.result)            # exposed property
```

## Test plan

- [ ] `pytest tests/core/test_core_command_nonblocking_construction.py -v` — construction, properties, error path
- [ ] `pytest tests/interfaces/test_interfaces_command_nonblocking.py -v` — ZMQ integration: blocking/non-blocking, re-entry, shutdown join, error path
- [ ] `pytest tests/interfaces/test_interfaces_simple_client_nonblocking.py -v` — SimpleClient exec + cookbook example smoke test
- [ ] `python -m pyrogue.examples._NonBlockingCommandExample` — cookbook example runs headless

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/src/pyrogue_tree/builtin_devices/index.rst
+++ b/docs/src/pyrogue_tree/builtin_devices/index.rst
@@ -59,7 +59,7 @@ Related Topics
 * Data receive/read path: :doc:`data_receiver`
 * Lower-level modules and wrappers: :doc:`/built_in_modules/index`
 * External process integration: :doc:`process`
-* Long-running operations: comparing pr.Process vs Command(nonBlocking=True): :doc:`long_running_operations`
+* Long-running operations: comparing pr.Process vs Command(blocking=False): :doc:`long_running_operations`
 
 .. toctree::
    :maxdepth: 1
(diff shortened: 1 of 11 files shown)


ruck314:
Closing this for now (reviewing whether the feature is needed before moving forward)

PR #1249, opened 2026-05-31
Revert changes from PR#1232

ruck314:
Reverts the ZmqServer._updateList concurrency guard and its slow-pickle race reproducer test introduced in PR#1232 (ESROGUE-711).


ruck314:
closing and going to revisit overly aggressive with locking

PR #1250, opened 2026-05-31
Revert changes from PR#1235

ruck314:
Reverts the VirtualClient bootstrap `ConnectionError`, the `_remoteAttr` liveness guard, the `ZmqClient::send` errno enrichment, and the two regression test files introduced by PR#1235, while preserving the unrelated PR#1238 daemon-monitor changes.


ruck314:
decided that we will keep this

Issue #1234, opened 2026-05-12
ZMQ client crashed with rogue 6.12.0


tristpinsm:
Observed trying to start a `pysmurf` client:

```
Connected to AMCc at localhost:9006
Traceback (most recent call last):
  File "/usr/local/src/rogue/python/pyrogue/interfaces/_Virtual.py", line 682, in _remoteAttr
    ret = pickle.loads(self._send(pickle.dumps({ 'path':path, 'attr':attr, 'args':args, 'kwargs':kwargs })))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
rogue.GeneralError: ZmqClient::send: General Error: zmq_sendmsg failed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/sodetlib/scripts/start_pysmurf_ipython.py", line 23, in <module>
    S = cfg.get_smurf_control(dump_configs=True)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sodetlib/sodetlib/det_config.py", line 687, in get_smurf_control
    S = pysmurf.client.SmurfControl(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/src/pysmurf/python/pysmurf/client/base/smurf_control.py", line 132, in __init__
    self.initialize(data_dir=data_dir,
  File "/usr/local/src/pysmurf/python/pysmurf/client/base/smurf_control.py", line 255, in initialize
    self.is_rfsoc = (True if 'Zcu' in self.get_fpga_build_stamp() else False)
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/src/pysmurf/python/pysmurf/client/command/smurf_command.py", line 3240, in get_fpga_build_stamp
    return self._caget(
           ^^^^^^^^^^^^
  File "/usr/local/src/pysmurf/python/pysmurf/client/command/smurf_command.py", line 190, in _caget
    var = self._client.root.getNode(pvname)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/src/rogue/python/pyrogue/interfaces/_Virtual.py", line 265, in _getNode
    obj = obj.node(a,load)
[...]

PR #1251, opened 2026-05-31
Revert changes from PR#1237

ruck314:
Reverts PR #1237 (the ZmqClient bounded-waitRetry retry cap), restoring the unbounded recv-retry contract while preserving the independent PR #1234 errno enrichment in `ZmqClient::send`.


ruck314:
We are going to close but update the API

PR #1256, opened 2026-06-02
fix(zmq): replace ZmqServer update lock with a lock-free deque

ruck314:
### Description

Replaces the `threading.Lock` added in #1232 (ESROGUE-711) for `ZmqServer`'s pending-update batching with a lock-free `collections.deque`. `_varUpdate()` now appends `(path, value)` (a single GIL-atomic op) and `_varDone()` drains the deque into a thread-private dict before serializing, so `pickle.dumps` only ever iterates a frame-local object no other thread can reach. This eliminates the `dictionary changed size during iteration` failure mode structurally — without taking a lock on the per-variable hot path — and keeps published payloads identical (last-write-wins per path, first-occurrence key order).

### Details

#1232 fixed a real defect: `_varDone()` pickled the **live** `_updateList` dict and only swapped it for an empty one *after* `pickle.dumps()` returned, so a value whose `__reduce__`/`__getstate__` releases the GIL (matplotlib transforms during SaveState/SaveConfig) let a concurrent caller mutate the dict mid-serialization. The fix wrapped both callbacks in a lock plus a snapshot-and-swap.

The concern raised on that merge was that locking `_varUpdate` is overly aggressive: it's the hot path (one call per variable per update group), and the ZMQ flush is already a known serialization point. In the shipped runtime the listener callbacks fire only from the single `Root._updateWorker` thread (`func` per variable, then `done`, sequentially), so the lock guards a cross-thread/direct-caller scenario rather than the normal path.

This change keeps the substantive guarantee from #1232 — *never serialize a shared object* — while making the hot path lock-free:

- `_varUpdate(path, value)` → `self._pendingUpdates.append((path, value))`
- `_varDone()` → drain via `popleft()` into a local `snapshot` dict, then `pickle.dumps(snapshot)` outside any lock.

`deque.append`/`popleft` are GIL-atomic (the same atomicity guarantee the original dict design relied on). A concurrent `_varUpdate` during a slow pickle simply appends to the deque and is published on the next flush.

`stream.interfaces.Variable._varDone()` is intentionally left untouched — PyYAML's `represent_mapping` snapshots via `list(mapping.items())`, so it was never exploitable (same scope decision as #1232).


slacrherbst:
I am curious. Does this pass the CI tests invented in the previous PR?

python/pyrogue/interfaces/_ZmqServer.py
```diff
@@ -108,17 +107,19 @@ def _doString(self, data: str) -> str:
 
     def _varUpdate(self, path: str, value: Any) -> None:
         """Queue a variable update for batching."""
-        with self._updateLock:
-            self._updateList[path] = value
+        self._pendingUpdates.append((path, value))
```
slacrherbst, line 110: So I don't think this is right. The previous version would use the dict with the path as the key to ensure only one entry per variable was kept in the queue. With this approach I think we end up with too many update events generated.

python/pyrogue/interfaces/_ZmqServer.py
```diff
@@ -108,17 +107,19 @@ def _doString(self, data: str) -> str:
...
-        self._publish(pickle.dumps(snapshot))
+        snapshot = {}
+        while True:
+            try:
+                path, value = self._pendingUpdates.popleft()
+            except IndexError:
+                break
+            snapshot[path] = value
+        if snapshot:
+            self._publish(pickle.dumps(snapshot))
```
slacrherbst, line 122: Ah it looks like this addresses the overlap on read. How is this for performance since we are looping through the entire variable of pending updates.

slacrherbst: Actually I think this might be ok as its on the update side which has less overlaps.

slacrherbst (approved):

python/pyrogue/interfaces/_ZmqServer.py
ruck314: The dedup is still there — `_varDone` drains the deque into a path-keyed dict, so repeated paths collapse last-write-wins to one entry per variable before publish, giving an identical payload to the old dict (the unit test asserts one entry per path).

python/pyrogue/interfaces/_ZmqServer.py
ruck314: Agreed — these callbacks fire sequentially from the single _updateWorker thread, so the deque stays short and the drain loop is negligible.

PR #1259, opened 2026-06-02
Serialize root operations with operation lock

bengineerd:
### Description

Add root-wide operation coordination so YAML configuration, bulk root reads and
writes, lifecycle resets, command execution, and ZMQ requests cannot interleave
with ordinary built-in variable access.

`Root` now owns a writer-preferred shared/exclusive operation gate. Normal
variable calls use shared admission and remain concurrent with one another,
while coordinated operations use reentrant exclusive admission. This closes the
original architecture's gap where EPICS or another direct variable entry point
could access the tree during an otherwise exclusive operation.

#### Architecture

- `Root.operationLock()` acquires the exclusive side of the gate. Exclusive
  owners may call variable APIs and nested root helpers reentrantly.
- Concrete local, remote, and link variable `get`, `set`, `post`, and `write`
  entry points acquire shared admission. Interfaces such as EPICS therefore
  participate without interface-specific locking.
- Command callbacks and direct remote-command writes acquire exclusive
  admission because commands commonly coordinate multiple variables.
- Existing readers may make nested shared calls while a writer waits, allowing
  callbacks and `LinkVariable` chains to finish. New top-level readers wait
  behind queued writers to prevent exclusive-operation starvation.
- Detached nodes continue to operate through no-op admission contexts.

The exclusive scope is applied to root read/write helpers, lifecycle helpers,
YAML load/set/export paths, remote variable/configuration dumps, and ZMQ request
handling. YAML helpers on detached nodes continue to raise an explicit
`NodeError` because they operate on a live rooted tree.

#### Current boundaries

- Shared-to-exclusive upgrades are intentionally unsupported. Code must enter
  `operationLock()` or an exclusive command before calling variables; attempting
  to acquire it from inside a shared variable callback raises `RuntimeError`.
- With `wait=False`, shared admission ends when the Python API method returns.
  Existing block and transaction locking governs any asynchronous hardware
  transaction still in flight.
- The gate covers Rogue's concrete built-in variable classes. A downstream
  `BaseVariable` subclass that completely replaces an access method can bypass
  admission unless it delegates to the built-in implementation.
- A local cached `LocalVariable.value()` measurement showed approximately one
  microsecond of added admission overhead per call. This is an implementation
  cost observation, not a maintained performance threshold.

#### Validation

- `17 passed` in `tests/core/test_core_root_device_io_edges.py`.
- `48 passed` across core variable/command, variable wait, YAML configuration,
  and ZMQ server tests.
- `git diff --check` passes.

### JIRA

https://jira.slac.stanford.edu/browse/ESROGUE-745

### Related

Related discussion: slaclab/rogue#1256


slacrherbst:
This appears to only create exclusivity between bulk read and write commands and the zmq interface. This would not block epics accesses, or other entry points that are not zmq. I can't seem to in my head figure out a clean way to do this that does not impact performance. the only way to do this is an attempted lock at each variable entry point which checks for a root level lock on a "exclusive operation" that includes the bulk configure and status. This lock would be re-entrant.

bengineerd:
I've updated this PR with a new architecture proposal. I'm still not sold on it though. We may just drop this entirely.

PR #1263, opened 2026-06-13
fix(memory): make Block getBytes/setBytes GIL release conditional

### LINKED ISSUE #1262: Rogue v6 update queue drain performance regression
I've been investigating an issue with the SO SMuRF software upgraded to use `rogue 6` where a call to `waitOnUpdate` following a specific operation (`take_bias_steps`) systematically stalls and causes the system to time out and crash. I solved (I think) the issue by changing the structure of the memory array that is accessed during that operation to more efficiently propagate through the rogue update thread in slaclab/amc-carrier-core#420. Given we did not observe as severe an issue with `rogue 4`, I suspect this may be pointing at a performance regression in rogue.

What follows is an analysis generated by Claude Code.

## Summary

We observed >120s stalls in `waitOnUpdate()` when the Root `_updateQueue` accumulated ~42k entries. The flood was caused by `LutMem.MemArray` register access without `updateGroup`, but the severity of the stall appears to be a v6 regression — the same register structures did not cause comparable stalls under rogue v4.

The root cause of the flood (missing `updateGroup` around `setArray`/`getArray`) is understood and being fixed in amc-carrier-core. This report focuses on why rogue v6 drains the queue so much more slowly than v4 once the queue is over-filled, which may point to latent performance issues.

## Primary issue: unconditional GIL release in Block::getBytes

The v6 C++ block layer releases the GIL unconditionally in `getBytes` (`Block.cpp:742`) and `setBytes` (`Block.cpp:658`):

```cpp
void rim::Block::getBytes(...) {
    rogue::GilRelease noGil;                    // always releases GIL
    std::lock_guard<std::mutex> lock(mtx_);     // acquires block mutex
    memcpy(data, blockData_ + offset, size);    // nanoseconds of actual work
}
```

In the update worker, each queue entry triggers `_doUpdate()` → `VariableValue()` → `var.get(read=False)` → C++ `getBytes`. With 42k entries, this produces 42k GIL release/reacquire cycles.

In v4, `RemoteBlock.get()` was pure Python (`_Block.py:370`). The worker held the GIL continuously while draining — no involuntary thread switches, no reacquisition delays.

Under thread contention (poll thread, hardware transaction threads, ZMQ server, EPICS), each GIL reacquisition can take 1–3ms, giving 42k × 1–3ms ≈ 42–126s. This matches our observed stall duration.

**Potential fix:** make the GIL release conditional — skip it when no hardware transaction is in progress and the call is just reading cached block data. The `read=False` path only needs an in-memory buffer read, so concurrency is unnecessary.

## Secondary issue: per-entry doneFunc overhead in the update worker

In v6, the update worker calls every registered listener's `doneFunc()` after processing each queue entry (`_Root.py:1203-1210`), regardless of whether any variables passed the listener's group filter:

```python
# Runs once per queue entry, even when no variables matched
with self._varListenLock:
    for func, doneFunc, incGroups, excGroups in self._varListeners:
        if doneFunc is not None:
            doneFunc()    # e.g. ZmqServer._varDone, stream.Variable._varDone
```

These calls are no-ops (early return) when no variables were added to the listener's update list, but the function call overhead, lock acquisition, and iteration still occur 42k times. In v4, the equivalent YAML/ZMQ work was hardcoded in the worker and could be structured per-batch rather than per-listener-per-entry.

This is a minor cost (~microseconds per entry), but it adds up at scale and is avoidable.

## Secondary issue: _recurseAddListeners per entry

The v6 worker calls `_recurseAddListeners()` for each variable in each queue entry (`_Root.py:1179-1180`). In v4, listener expansion was done on the producer side inside `_queueUpdate()`. For variables with no listeners this is a no-op, but it's additional per-entry work that v4 didn't have in the consumer path.

## Secondary issue: _isSystemLogVar check per entry

Each queue entry evaluates `all(self._isSystemLogVar(v) for v in uvars.values())` (`_Root.py:1168`). For non-syslog variables this is a fast identity check, but it's another per-entry cost that v4 didn't have.

## Context

- Rogue version: v6.13.0 (comparison baseline: v4.11.x)
- Queue contents: ~42k entries, predominantly `LutCtrl.Lut[*].MEM[*]` registers (not in stream group, so no YAML serialization per entry)
- Trigger: `take_bias_steps` in sodetlib calling `setArray`/`getArray` on `LutMem.MemArray` without `updateGroup`
- Queue dump available in `q_dump.json` (42,020 entries, 41,620 from LutCtrl)

## Reproducer

Any code path that sets a large number of variables without `updateGroup` will expose the GIL-cycling overhead. The LUT register access is the most acute case because `numValues=2048` per channel × 2 channels × multiple accesses ≈ 40k+ individual updates.

ruck314:
### Description
Fixes a v6 performance regression (#1262) where the Root update-queue drain stalls for tens of seconds. `Block::getBytes` and `Block::setBytes` released the Python GIL unconditionally on every call; a bulk drain of ~42k staged variable updates therefore performed ~42k GIL release/re-acquire cycles, and under GIL contention each hand-off costs milliseconds. The method bodies are pure in-memory work (memcpy/copyBits/byte-reverse) with no Python calls, so the GIL now stays held on the uncontended fast path and is released only while blocking on a contended block mutex, preserving the deadlock-avoidance contract with memory-transaction workers. Contended 42k-entry drain improves from ~16 s to ~0.05 s in the reproduction.

### Details
- `Block::setBytes` / `Block::getBytes` now acquire `mtx_` with `std::unique_lock(mtx_, std::try_to_lock)`. On the uncontended path the GIL is never dropped (kills the per-call thrash). Only when the lock is contended -- the case that could deadlock against a transaction worker holding `mtx_` while needing the GIL for a Python callback -- is `rogue::GilRelease` used while blocking on `lock.lock()`. `intStartTransaction` / `checkTransaction` (which genuinely block via `waitTransaction`) are intentionally left unchanged.
- Reproduction: `tests/perf/test_block_gil_contention_perf.py` (marked `perf`) drives a 42k staged `set(write=False)`/`get(read=False)` drain with and without a background GIL-contending thread. Pre-fix the contended drain trips the ceiling (~16 s, ~517x slowdown); post-fix it stays sub-second.
- Correctness: `tests/core/test_block_gil_fastpath_roundtrip.py` pins staged set/get round-trip equivalence on the scalar, list, and byte-reversed (malloc-copy) paths, single-threaded and under same-Block contention (exercising the `try_to_lock` fallback branch).
- Verified locally: full Python suite passes (one unrelated pre-existing PyDM/PyQt test failure), `ctest -L cpp` 15/15, no-Python small build `ctest -L no-python` 12/12, `check_whitespace.sh` and `run_linters.sh` clean.
- Secondary per-entry overheads in `_Root.py::_updateWorker` noted in #1262 are left out of scope to keep the change surgical; they touch listener fan-out semantics and warrant their own change + tests.

### JIRA
N/A

### Related
Closes #1262

### DIFF
--- a/src/rogue/interfaces/memory/Block.cpp
+++ b/src/rogue/interfaces/memory/Block.cpp
@@ -38,6 +38,7 @@ namespace bp = boost::python;
 #include <exception>
 #include <iomanip>
 #include <memory>
+#include <mutex>
 #include <sstream>
 #include <string>
 #include <vector>
@@ -655,8 +656,18 @@ void rim::Block::setBytes(const uint8_t* data, rim::Variable* var, uint32_t inde
     uint8_t tmp;
     uint8_t* buff;
 
-    rogue::GilRelease noGil;
-    std::lock_guard<std::mutex> lock(mtx_);
+    // Fast path: take mtx_ without dropping the GIL. The body below is pure
+    // in-memory work (memcpy/copyBits/byte-reverse) with no Python calls, so
+    // holding the GIL across it is correct and avoids the release/re-acquire
+    // thrash that dominates bulk update-queue drains (SLAC rogue #1262). Only
+    // when mtx_ is actually contended -- held by a memory-transaction worker
+    // that may need the GIL for a Python callback -- do we drop the GIL while
+    // blocking, preserving the deadlock-avoidance contract.
+    std::unique_lock<std::mutex> lock(mtx_, std::try_to_lock);
+    if (!lock.owns_lock()) {
+        rogue::GilRelease noGil;
+        lock.lock();
+    }
 
     // Set stale flag
     if (var->mode_ != "RO") stale_ = true;
@@ -739,8 +750,15 @@ void rim::Block::getBytes(uint8_t* data, rim::Variable* var, uint32_t index) {
     uint32_t dstBit;
     uint32_t x;
 
-    rogue::GilRelease noGil;
-    std::lock_guard<std::mutex> lock(mtx_);
+    // Fast path: take mtx_ without dropping the GIL. See setBytes() for the full
+    // rationale -- the body is pure in-memory work, so we only release the GIL
+    // on the contended-wait path to preserve deadlock avoidance (SLAC rogue
+    // #1262).
+    std::unique_lock<std::mutex> lock(mtx_, std::try_to_lock);
+    if (!lock.owns_lock()) {
+        rogue::GilRelease noGil;
+        lock.lock();
+    }
 
     // List variable
     if (var->numValues_ != 0) {
--- a/tests/core/test_block_gil_fastpath_roundtrip.py
+++ b/tests/core/test_block_gil_fastpath_roundtrip.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+#-----------------------------------------------------------------------------
+# Title      : Block getBytes/setBytes conditional-GIL fast-path tests
+#-----------------------------------------------------------------------------
+# This file is part of the rogue software platform. It is subject to
+# the license terms in the LICENSE.txt file found in the top-level directory
+# of this distribution and at:
+#    https://confluence.slac.stanford.edu/display/ppareg/LICENSE.html.
+# No part of the rogue software platform, including this file, may be
+# copied, modified, propagated, or distributed except according to the terms
+# contained in the LICENSE.txt file.
+#-----------------------------------------------------------------------------
+#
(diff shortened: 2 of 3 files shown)


slacrherbst:
Ok. This fix appears valid. I wonder if there are other areas where we can use this try lock, and only gil release if we are waiting on lock approach.
@bengineerd

slacrherbst (approved):

tristpinsm:
I confirmed that the very slow draining of the queue reported in #1262 goes away with this fix.

PR #1267, opened 2026-06-16
Add configurable UpdatePeriod for updateGroup leak period

### LINKED ISSUE #1266: `LoadConfigProcess` is not isolated from variable listener updates
A `Process` runs its task within a `updateGroup(period=1.0)` context, which means that variable updates can proceed and trigger listeners. This causes conflicts when setting up devices in `pysmurf`.

Possible solutions are to make the period configurable or override the `_run` method in these cases.

slacrherbst:
## Summary
  
- Adds a hidden `RW` `LocalVariable` named `UpdatePeriod` (float, default `1.0s`) to the base `Process` class
- Uses `self.UpdatePeriod.value()` as the `period` argument to `updateGroup()` in `_run()`, replacing the previously hardcoded `1.0` value
- All `Process` subclasses (including `LoadConfigProcess`, `SaveConfigProcess`, `RunControl`, etc.) inherit this variable automatically
  
Closes #1266

### DIFF
--- a/docs/src/pyrogue_tree/builtin_devices/process.rst
+++ b/docs/src/pyrogue_tree/builtin_devices/process.rst
@@ -76,6 +76,7 @@ The built-in status Variables are:
 * ``Progress`` for fractional completion
 * ``Message`` for operator-facing status text
 * ``Step`` and ``TotalSteps`` for step-count style progress reporting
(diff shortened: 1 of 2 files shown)


ruck314 (approved):
@slacrherbst LGTM. Merge when ready

PR #1276, opened 2026-07-02
fix(rocev2.Server): free RX slab in destructor to fix zero-copy teardown segfault

ruck314:
### Description

Fix a use-after-free in `rogue::protocols::rocev2::Server` teardown. `stop()` freed the registered RX slab while zero-copy `Buffer`s already handed downstream still pointed into it, which surfaced as a segmentation fault during process or GUI teardown.

The slab is now the `Pool`'s buffer backing and lives until `~Server()`; `stop()` releases only the external ibverbs resources. This applies the same shutdown contract that landed for `AxiStreamDma` in #1277.

### Changes

- `~Server()` frees the RX slab; `stop()` destroys only the QP/CQ/comp-channel and deregisters the MR. Each zero-copy `Buffer` holds a `shared_ptr` to its `Pool`, so the backing outlives every outstanding frame. The failed-construction path still frees the slab explicitly.
- `resourcesMtx_` serializes `postRecvWr()` against `cleanupResources()`. The QP/MR validity check and `ibv_post_recv()` run inside that lock, so a deferred return either re-posts before teardown or observes the resources are gone. `retBuffer()` no longer reads `threadEn_`/`qp_` outside the lock.
- `completeConnection()` rejects use after `stop()` with a lifecycle error before any ibverbs call.
- Adds missing `GilRelease` coverage around ibverbs bring-up, connection completion, deferred returns, and shutdown (`stop()` joins a receive thread that re-acquires the GIL).
- Hardware-gated `rxe0` regression covering retained-frame validity across `stop()`, concurrent returns versus `stop()`, the post-stop `completeConnection()` guard, `stop()` idempotency, and `weak_ptr`-verified destruction.

### Lifecycle contract

Callers must quiesce `completeConnection()` and other ibverbs-resource operations before calling `stop()`. Returns of previously issued zero-copy buffers are the only operation allowed to overlap shutdown.

### Validation

- Linux build clean; `ctest -L cpp` 18/18 pass; `cpplint` and `check_whitespace.sh` clean.
- Teardown regression run against a real RoCEv2 NIC: 22/22 assertions pass. Reverting the slab-lifetime fix makes it crash with SIGSEGV at the post-teardown read; removing the `completeConnection()` entry guard makes the error surface from `postRecvWr()` instead and fails the new assertion.
- Earlier end-to-end validation against a 10 GbE RoCEv2 firmware target: 25/25 PRBS runs with no teardown segfault.

### Known limitation

The return-versus-`stop()` regression starts both operations from a rendezvous and returns several buffers, but it cannot force a specific interleaving inside `libibverbs` the way #1277's fake-driver shim can. Each active zero-copy return also takes one extra mutex before the existing `decCounter()` lock; there is no focused RoCE return-throughput measurement in this PR.

### JIRA

N/A

### DIFF
--- a/src/rogue/protocols/rocev2/Server.cpp
+++ b/src/rogue/protocols/rocev2/Server.cpp
@@ -770,6 +770,11 @@ void rpr::Server::acceptFrame(ris::FramePtr /*frame*/) {
 // stop / destructor
 // ---------------------------------------------------------------------------
 void rpr::Server::stop() {
+    // Callers must quiesce completeConnection() and other public ibverbs
+    // resource operations before shutdown.  Deferred zero-copy buffer returns
+    // are the one operation allowed to overlap stop(); resourcesMtx_ serializes
+    // their receive-WR re-posts with cleanupResources() below.
+
     // Release the GIL for the duration of teardown.  stop() is driven from Python
     // (RoCEv2Server._stop) with the GIL held, and the receive thread re-acquires
     // the GIL via ScopedGil inside sendFrame() when a downstream slave is Python;
--- a/src/rogue/protocols/rocev2/Server.cpp
+++ b/src/rogue/protocols/rocev2/Server.cpp
@@ -311,8 +311,13 @@ rpr::Server::Server(const std::string& deviceName,
 // Buffer still held downstream (each points into slab_ and keeps this Server
 // alive via shared_ptr<Pool>).  ~Server frees the slab; the failed-construction
 // path frees it explicitly (the destructor does not run on a partial object).
+//
+// Serialized with postRecvWr() so a deferred retBuffer() cannot re-post using
+// qp_/mr_ while stop() is destroying them.
 // ---------------------------------------------------------------------------
 void rpr::Server::cleanupResources() {
+    std::lock_guard<std::mutex> lock(resourcesMtx_);
+
     if (qp_) {
         ibv_destroy_qp(qp_);
         qp_ = nullptr;
@@ -478,6 +483,8 @@ std::string rpr::Server::getGid() const {
 // wr_id == slot index so no lookup is needed on completion
 // ---------------------------------------------------------------------------
 void rpr::Server::postRecvWr(uint32_t slot) {
+    std::lock_guard<std::mutex> lock(resourcesMtx_);
+
     // Defensive: slot indexes into slab_; a corrupted wr_id (from the CQ)
     // or meta (from retBuffer) must not produce an out-of-bounds slab
     // pointer.  Normal control flow keeps slot < numBufs_ because we set
@@ -487,6 +494,11 @@ void rpr::Server::postRecvWr(uint32_t slot) {
         throw(rogue::GeneralError::create("rocev2::Server::postRecvWr",
                                           "slot=%u out of range (numBufs=%u)",
                                           slot, numBufs_));
+
+    if (!qp_ || !mr_ || !slab_)
+        throw(rogue::GeneralError::create("rocev2::Server::postRecvWr",
+                                          "resources are not available for slot=%u",
+                                          slot));
 
     uint8_t* bufStart = slab_ + (static_cast<uint64_t>(slot) * bufSize_);
 
@@ -528,7 +540,7 @@ void rpr::Server::retBuffer(uint8_t* data, uint32_t meta, uint32_t rawSize) {
 
     log_->debug("retBuffer: re-posting slot=%u", slot);
 
-    if (threadEn_.load() && qp_) {
+    if (threadEn_.load()) {
         try {
             postRecvWr(slot);
         } catch (...) {
@@ -786,13 +798,14 @@ void rpr::Server::stop() {
         thread_ = nullptr;
     }
     // Release the external ibverbs resources now — QP/CQ/MR(dereg)/comp-channel
-    // and the wake pipe — mirroring AxiStreamDma::stop(), which releases its
-    // external kernel resources (DMA mmap + fd) at stop().  But NOT slab_: the RX
-    // slab is this Pool's buffer backing, and zero-copy Buffers handed downstream
-    // still point into it.  Per the base Pool contract (ris::Pool::~Pool() frees
-    // its buffer memory at destruction, and every Buffer holds a shared_ptr to
-    // its Pool), the slab's lifetime is the Server object's; it is freed in
-    // ~Server (below), after the last outstanding Buffer has released the Server.
+    // and the wake pipe — mirroring AxiStreamDma::stop(), which closes its
+    // per-instance fd while leaving zero-copy backing memory alive until
+    // destruction.  The RX slab is this Pool's buffer backing, and zero-copy
+    // Buffers handed downstream still point into it.  Per the base Pool contract
+    // (ris::Pool::~Pool() frees its buffer memory at destruction, and every
+    // Buffer holds a shared_ptr to its Pool), the slab's lifetime is the Server
+    // object's; it is freed in ~Server (below), after the last outstanding Buffer
+    // has released the Server.
     cleanupResources();
 }
 
--- a/src/rogue/protocols/rocev2/Server.cpp
+++ b/src/rogue/protocols/rocev2/Server.cpp
@@ -102,6 +102,13 @@ rpr::Server::Server(const std::string& deviceName,
     memset(fpgaGid_, 0, 16);
     wakeFd_[0] = wakeFd_[1] = -1;
 
+    // Release the GIL for the ibverbs bring-up below (open device, reg_mr, create
+    // CQ/QP, modify to INIT, query GID) — blocking syscalls run while the Python
+    // caller holds the GIL.  Matches AxiStreamDma's constructor.  On a throw,
+    // GilRelease's destructor re-acquires the GIL before the boost.python
+    // exception translator runs.
+    rogue::GilRelease noGil;
+
     // The destructor does NOT run on a partially-constructed object, so any
     // throw between here and the end of the body would leak slab_ / mr_ /
     // cq_ / qp_.  Wrap the body in try/catch and call cleanupResources() on
@@ -281,15 +288,29 @@ rpr::Server::Server(const std::string& deviceName,
         log_->info("RC QP ready: qpn=0x%06x rqPsn=0x%06x sqPsn=0x%06x",
                    hostQpn_, hostRqPsn_, hostSqPsn_);
     } catch (...) {
+        // Failed construction: the destructor does NOT run on a partially
+        // constructed object, so release the ibverbs resources and free the slab
+        // here.  No zero-copy Buffer exists yet, so freeing the slab is safe.
         cleanupResources();
+        if (slab_) {
+            free(slab_);
+            slab_ = nullptr;
+        }
         throw;
     }
 }
 
 // ---------------------------------------------------------------------------
-// cleanupResources — release every ibverbs / heap resource owned by Server,
-// in reverse order of allocation.  Idempotent (safe to call from both the
-// failed-construction path and stop()).
+// cleanupResources — release the EXTERNAL ibverbs resources owned by Server
+// (QP, CQ, comp-channel, MR registration, wake pipe), in reverse order of
+// allocation.  Idempotent.
+//
+// The RX slab is deliberately NOT freed here: it is this Pool's buffer backing,
+// and per the base Pool convention (ris::Pool::~Pool) its lifetime is the Server
+// object's.  Freeing it in stop() would be a use-after-free for any zero-copy
+// Buffer still held downstream (each points into slab_ and keeps this Server
+// alive via shared_ptr<Pool>).  ~Server frees the slab; the failed-construction
+// path frees it explicitly (the destructor does not run on a partial object).
 // ---------------------------------------------------------------------------
 void rpr::Server::cleanupResources() {
     if (qp_) {
@@ -308,10 +329,8 @@ void rpr::Server::cleanupResources() {
         ibv_dereg_mr(mr_);
         mr_ = nullptr;
     }
-    if (slab_) {
-        free(slab_);
-        slab_ = nullptr;
-    }
+    // slab_ is intentionally NOT freed here — it is the Pool buffer backing (see
+    // the comment above); ~Server and the failed-construction path free it.
     for (int i = 0; i < 2; ++i) {
         if (wakeFd_[i] >= 0) {
             close(wakeFd_[i]);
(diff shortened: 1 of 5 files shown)


tests/cpp/protocols/rocev2/test_rocev2_teardown.cpp
```diff
@@ -0,0 +1,382 @@
...
+    ibv_context* ctx = nullptr;
+    ibv_pd*      pd  = nullptr;
+    ibv_cq*      cq  = nullptr;
+    ibv_qp*      qp  = nullptr;
+    ibv_mr*      mr  = nullptr;
+    std::vector<uint8_t> sbuf;
+    ibv_gid  gid{};
+    uint8_t  gidIndex = 0;
+
+    bool open() {
```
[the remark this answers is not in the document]
ruck314: Valid — agreed. \`doctest\` \`REQUIRE()\` throws on failure, and any \`REQUIRE()\` between \`client.open()\` and the trailing \`client.destroy()\` would unwind the stack while \`LoopbackClient\` had no destructor, leaking \`ctx/pd/cq/qp/mr\`. Fixed in 1ffa469: added \`~LoopbackClient()\` that calls the already-idempotent \`destroy()\`, and deleted the copy ctor/assignment so the raw owning handles cannot be double-freed. The explicit \`client.destroy()\` at the end is now a safe no-op before the destructor runs. Rebuilt and re-ran the rxe0 loopback test (12/12 assertions pass).

tests/cpp/protocols/rocev2/test_rocev2_teardown.cpp
```diff
@@ -0,0 +1,382 @@
...
+    // Server::stop() -> cleanupResources() free()s the slab that `frame` still
+    // points into.  (Fixed build: the slab is freed only in ~Server, which the
+    // frame's shared_ptr<Pool> keeps alive, so it outlives the frame.)
+    server->stop();
+
+    // Reclaim the freed region: many same-size allocations, each fully written,
+    // so the allocator hands the just-freed slab chunk back and we overwrite it.
+    // (free() of this multi-hundred-KiB slab returns it to the glibc arena rather
+    // than munmap'ing, so the stale bytes survive until something reuses them —
+    // hence this deterministic clobber instead of relying on a fault.)  On the
```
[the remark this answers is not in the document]
ruck314: Valid — the two comment blocks did contradict each other (one said the mmap-backed slab is munmap'd by \`free()\` so the read faults; the other said \`free()\` returns it to the glibc arena rather than munmap'ing). The real behavior is allocator- and threshold-dependent (glibc's dynamic mmap threshold rises after a large mmap'd block is freed, so same-size reallocations come from the arena; other allocators differ), and the test deliberately does not depend on a fault. Fixed in 1ffa469: both blocks now say the slab is large enough to be mmap-backed, that whether \`free()\` munmaps or returns to the arena is allocator/threshold-dependent, and that the clobber is a best-effort provoke of reuse so the failure is a deterministic byte mismatch when it does not fault outright.

tests/cpp/protocols/rocev2/test_rocev2_teardown.cpp
```diff
@@ -0,0 +1,395 @@
...
+int findIpv4GidIndex(ibv_context* ctx, uint8_t port) {
+    for (int i = 0; i < 16; ++i) {
+        ibv_gid g;
+        if (ibv_query_gid(ctx, port, i, &g) != 0) continue;
+        if (g.raw[10] == 0xff && g.raw[11] == 0xff &&
+            (g.raw[12] | g.raw[13] | g.raw[14] | g.raw[15]) != 0)
+            return i;
+    }
+    return -1;
+}
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in 4a53900. Two real defects, confirmed on this host's `rxe0` (`gid_tbl_len=1024`, IPv4-mapped GID at index 1, `fe80::` link-local at index 0):

1. **Hard-capped scan.** The loop stopped at 16, but as `Server.cpp` itself notes, `rdma_rxe` advertises `gid_tbl_len=1024`; the IPv4-mapped RoCEv2 GID is not guaranteed to sit in the first 16 slots. `findIpv4GidIndex` now `ibv_query_port`s the port and scans the full `pattr.gid_tbl_len`.
2. **Partial prefix match.** Checking only `bytes[10..11]==0xffff` plus a non-zero tail would also accept a non-IPv4-mapped GID whose leading bytes happen to be non-zero. The predicate now requires the complete `::ffff:0:0/96` prefix — `bytes[0..9]==0` **and** `bytes[10..11]==0xffff` — before returning the index.

Rebuilt and re-ran the `rxe0` loopback test: the probe resolves index 1 and all 12 assertions pass. `cpplint` clean.

tests/cpp/protocols/rocev2/test_rocev2_teardown.cpp
```diff
@@ -0,0 +1,395 @@
...
+    // A premature free() may munmap the slab (a later read then faults) or return
+    // it to the allocator arena (the stale bytes linger until reused) depending on
+    // the allocator and its mmap threshold; this clobber is a best-effort provoke
+    // of reuse so the failure is a deterministic byte mismatch when it does not
+    // fault outright.  On the fixed build the slab is still live, so these land
+    // elsewhere and the frame is untouched.
+    const size_t slabBytes = static_cast<size_t>(kMaxPayload) * kQueueDepth;
+    std::vector<std::vector<uint8_t>> clobber;
+    for (int i = 0; i < 64; ++i)
+        clobber.emplace_back(slabBytes, 0xAA);
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in 4a53900: added `clobber.reserve(64)` before the loop so the outer vector doesn't regrow.

One nuance on the stated cost: the 64 inner buffers are `std::vector<uint8_t>` handles, so on regrowth they're *moved* (three-pointer swaps), not deep-copied — the large slab-sized allocations aren't re-copied. So the avoided overhead is the outer-vector reallocation and pointer churn rather than re-allocating megabytes, but reserving is a strict, harmless improvement and makes the intent explicit. Re-ran the `rxe0` teardown test (12/12 assertions pass).

tests/cpp/protocols/rocev2/test_rocev2_teardown.cpp
```diff
@@ -0,0 +1,412 @@
...
+    // Use the IPv4-mapped RoCEv2 GID (skip the link-local fe80:: at index 0,
+    // which rxe rejects at INIT->RTR). Same index for the Server and client.
+    const int gidIdx = findIpv4GidIndex(client.ctx, 1);
+    REQUIRE(gidIdx >= 0);
+    REQUIRE(client.queryGid(static_cast<uint8_t>(gidIdx)));
+
+    // Server (receiver) on rxe0 at the IPv4-mapped GID index.
+    auto server = rpr::Server::create(kDev, 1, static_cast<uint8_t>(gidIdx),
+                                      kMaxPayload, kQueueDepth);
+    REQUIRE(server != nullptr);
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in cb093cd. `findIpv4GidIndex` scans the full `gid_tbl_len` (confirmed `1024` on this host's `rxe0`) and returns an `int`, but `gidIdx` was `static_cast<uint8_t>`'d twice — for `queryGid()` and `Server::create()`. Both the ibverbs `ah_attr.grh.sgid_index` field and `Server::create`'s `gidIndex` parameter are `uint8_t`, so an IPv4-mapped GID landing at an index > 255 would silently truncate (e.g. 300 → 44) and test the wrong GID index. Added `REQUIRE(gidIdx <= 255)` and now reuse a single `uint8_t gidIndex` for both call sites, so any out-of-range index fails loudly instead of silently probing the wrong slot. Rebuilt and re-ran the `rxe0` loopback test — the probe resolves index 1 and all assertions pass (13/13 now, the extra one being the bound check). `cpplint` clean.

tests/cpp/protocols/rocev2/test_rocev2_teardown.cpp
```diff
@@ -0,0 +1,416 @@
...
+    // Buggy build: use-after-free -> bytes are the 0xAA clobber (or a fault).
+    std::vector<uint8_t> after(kPayloadLen, 0);
+    {
+        auto it = frame->beginRead();
+        ris::fromFrame(it, kPayloadLen, after.data());
+    }
+    CHECK_EQ(std::memcmp(after.data(), client.sbuf.data(), kPayloadLen), 0);
+
+    // Release the held frame before the client MR/QP it (indirectly) races.
+    frame.reset();
```
bengineerd, line 511: This does not actually release the last held frame. `server` owns `slave` via `addSlave()`, `HoldingSlave::frames_` owns the `FramePtr`, and that frame's `Buffer` owns a `shared_ptr` back to the `Server`. `frame.reset()` only drops the local copy, so the test exits with a `Server -> slave -> frame -> buffer -> Server` cycle and `~Server()` never frees the slab. Please add a `HoldingSlave::clear()` (or similar) and call it here before leaving the test.

src/rogue/protocols/rocev2/Server.cpp
```diff
@@ -495,6 +519,11 @@ void rpr::Server::postRecvWr(uint32_t slot) {
 // meta lower 24 bits = slot index (set in createBuffer() call in runThread)
 // ---------------------------------------------------------------------------
 void rpr::Server::retBuffer(uint8_t* data, uint32_t meta, uint32_t rawSize) {
+    // retBuffer runs from Buffer::~Buffer(), which can fire on a Python thread
+    // holding the GIL; release it around the ibv_post_recv re-post and decCounter
+    // (which locks).  Matches ris::Pool::retBuffer() and AxiStreamDma::retBuffer().
+    rogue::GilRelease noGil;
```
bengineerd, line 537: The re-post check just below this is still racing teardown. A downstream frame can be released on another thread while `stop()` is past `threadEn_.store(false)` and about to destroy `qp_`/deregister `mr_` in `cleanupResources()`. Since `retBuffer()` reads `threadEn_` and `qp_` without synchronizing with `cleanupResources()`, it can enter `postRecvWr()` with a QP/MR being destroyed. The re-post path and resource cleanup need shared exclusion (for example, a mutex around the `threadEn_/qp_/mr_` check+post and `cleanupResources()`), or teardown needs another way to wait out active `retBuffer()` calls.

bengineerd (commented):
Review findings from local inspection of PR head 4fa2644.

src/rogue/protocols/rocev2/Server.cpp
```diff
@@ -749,10 +785,31 @@ void rpr::Server::stop() {
         delete thread_;
         thread_ = nullptr;
     }
+    // Release the external ibverbs resources now — QP/CQ/MR(dereg)/comp-channel
+    // and the wake pipe — mirroring AxiStreamDma::stop(), which releases its
```
bengineerd: Update: I split the AxiStreamDma side of this into #1277.

I would not use `AxiStreamDma::stop()` as the lifecycle precedent here without caveating it. Looking at `AxiStreamDma`, it appears to have the same latent issues: RX zero-copy frames are created over `desc_->rawBuff[...]`, `stop()` can call `closeShared(desc_)` and `dmaUnMapDma()` that mapping while downstream frames may still hold those buffers, and `retBuffer()` checks `fd_ >= 0` before `dmaRetIndex(fd_, ...)` without synchronizing against `stop()` closing `fd_`. So matching the old AxiStreamDma behavior does not prove this teardown contract is safe; the AxiStreamDma fix is now tracked separately in #1277, and this PR should avoid presenting the old AxiStreamDma behavior as a safe model.

bengineerd (commented):
Additional review note on the AxiStreamDma teardown comparison. The AxiStreamDma follow-up is now split into #1277.

src/rogue/protocols/rocev2/Server.cpp
ruck314: Closed in f475657. `resourcesMtx_` now covers both sides of the window you described: `postRecvWr()` takes it and validates `qp_`/`mr_`/`slab_` **inside** the lock, immediately before `ibv_post_recv()`, and `cleanupResources()` holds the same lock across the QP destroy and MR dereg. A deferred return therefore either completes its re-post before teardown starts or observes the resources are already gone; it can never enter `ibv_post_recv()` on a QP being destroyed.

This latest commit also removes the `threadEn_.load() && qp_` pre-check entirely. That read sat outside the mutex and was not what made the re-post safe, so leaving it in only obscured which check is load-bearing. `retBuffer()` now calls `postRecvWr()` unconditionally and lets the locked check decide, which is the same shape as `AxiStreamDma::retBuffer()` after #1277 (validate the descriptor inside the lock `stop()` takes to close it).

tests/cpp/protocols/rocev2/test_rocev2_teardown.cpp
ruck314: Addressed. `HoldingSlave` gained `takeAll()` and `clear()`, and the test now detaches every retained frame from the slave with `takeAll()` before teardown, so the `Server -> slave -> frame -> buffer -> Server` cycle is broken while the explicit `FramePtr`s needed for the lifetime and teardown-race checks are kept in a local vector.

Two related additions:

- A `HoldingSlaveCleanup` RAII guard stops delivery and then clears the slave, so a fatal `REQUIRE()` unwinding mid-test cannot leave the cycle in place or let an arriving frame recreate it.
- The test ends with `std::weak_ptr<Server>` and `CHECK(weakServer.expired())` after `frame.reset()` / `server.reset()`, which fails loudly if any ownership cycle survives.

Verified end to end on a real RoCEv2 NIC in this environment (22/22 assertions pass), including the negative case: restoring the slab free inside `cleanupResources()` makes the test crash with SIGSEGV at the post-teardown read.

src/rogue/protocols/rocev2/Server.cpp
ruck314: Fair point, and it is fixed in f475657. The `stop()` comment no longer uses AxiStreamDma as the justification. It now derives the rule from the base Pool contract directly (`ris::Pool::~Pool()` frees its buffer backing at destruction, and every `Buffer` holds a `shared_ptr` to its Pool, so the slab's lifetime is the Server object's), and only mentions AxiStreamDma as a parallel now that #1277 has landed the same split there.

I also brought the two #1277 code fixes across:

- `completeConnection()` gained an entry guard that rejects use after `stop()` with a lifecycle error, mirroring `acceptReq()`/`acceptFrame()`. Previously the call fell through to the recv-WR pre-post loop and surfaced the failure from `postRecvWr()` after ibverbs work had already been attempted on a destroyed QP/MR.
- `retBuffer()` no longer reads `threadEn_`/`qp_` outside `resourcesMtx_` (see the other thread).

bengineerd:
Addressed the GIL-scope issue in commit 70032f1e5.

The Server constructor body cannot release the GIL for device open because its Core base constructor has already completed by then. Core::Core now owns a GilRelease guard around device discovery, ibv_open_device(), and ibv_alloc_pd(). Core::~Core also releases the GIL around ibv_dealloc_pd() and ibv_close_device(), since Server::stop() has returned and reacquired the GIL before the base destructor runs.

The existing Server-level guard remains in place for the remaining MR/CQ/QP bring-up, and its comment now describes that scope accurately. This follows the same resource-boundary structure used by AxiStreamDma in PR 1277.

Validation: the Miniforge rogue_build build, whitespace check, and git diff --check pass. The local macOS build disables Linux-only RoCEv2 support, so compilation/execution of the affected RoCEv2 path remains covered by Linux CI.

bengineerd (approved):

PR #1283, opened 2026-08-11
Fix YAML serialization of display-formatted floats

### LINKED ISSUE #1282: FileReader config decode fails: DataWriter.Bandwidth disp='{:,.3f}' emits unparseable !!float with commas
## Summary

`FileReader(configChan=...)` fails to decode the config/status frame written by `StreamWriter`/`DataWriter` because the built-in `DataWriter.Bandwidth` variable is serialized using its **display formatter** `disp='{:,.3f}'`, which inserts comma thousands-separators. The value is emitted as a `!!float` scalar (e.g. `!!float '256,939.708'`), which PyYAML then cannot parse back:

```
ValueError: could not convert string to float: '256,939.708'
```

A single unparseable leaf aborts the entire `configDict` decode, so **no** config channel written by a `DataWriter` with nonzero bandwidth can be read back.

## Root cause

`python/pyrogue/_DataWriter.py` defines:

```python
self.add(pr.LocalVariable(
    name='Bandwidth',
    ...
    typeStr='Float64',
    localGet=self._getBandwidth,
    units='Bytes/sec',
    disp='{:,.3f}',   # <-- thousands-separator display format
    description='Instantaneous write bandwidth in bytes per second.'))
```

The config serialization writes the **disp-formatted string** as the `!!float` value. `disp` is a display concern and should not affect the serialized value; the comma grouping produces YAML that is not round-trippable.

Present at current HEAD (verified past v6.15.0).

## Minimal reproduction

```python
import pyrogue, pyrogue.utilities.fileio as fio, pyrogue.interfaces.stream, tempfile
root = pyrogue.Root(name='root', pollEn=False, initRead=False)
cs = pyrogue.interfaces.stream.Variable(root=root)
root.add(fio.StreamWriter(name='DataWriter', configStream={255: cs}))
root.start()
root.DataWriter.Bandwidth.set(335610724.312)   # any large bandwidth
fn = tempfile.mktemp(suffix='.dat')
root.DataWriter.DataFile.set(fn)
root.DataWriter.Open(); root.DataWriter.Close()
root.stop()
with fio.FileReader(files=[fn], configChan=255) as fd:
    for h, d in fd.records(): pass
    print(fd.configDict)   # -> ValueError: could not convert string to float: '...'
```

## Environment

- rogue 6.15.0
- PyYAML 6.0.3
- Python 3.13.15

## Suggested fix

Either drop the comma from the `Bandwidth` `disp` (`'{:.3f}'`), or — more generally — ensure numeric variables are serialized from their native value rather than the `disp`-formatted string, so display formatting can never break config round-tripping.

bengineerd:
### Description

Serialize PyRogue float `VariableValue` objects from their native values rather than their display-formatted strings. This keeps configuration and status YAML round-trippable when a float display format contains thousands separators, fixing `FileReader(configChan=...)` failures for `DataWriter.Bandwidth`.

### Details

`DataWriter.Bandwidth` uses `disp='{:,.3f}'`. The serializer previously tagged the display text as `!!float`, producing values such as `!!float '256,939.708'` that PyYAML could not parse. Float values now use PyYAML's native float representation, while boolean, enum, integer, and string display behavior remains unchanged.

Native float serialization also preserves full precision rather than the rounded display value. This is the intended round-trip behavior, but downstream golden configuration files may show expected float-value changes.

NumPy float scalars are converted to Python floats before representation so the YAML remains portable and contains no Python-specific NumPy object tags.

Integer display text remains unchanged to preserve readable formats such as hexadecimal register values. Safe handling of non-YAML integer display formats such as comma-grouped decimal output is tracked in #1284.

A regression test verifies that grouped float display formatting is preserved for the UI, omitted from persisted YAML, and decoded by `FileReader` as a native float. It also covers `np.float64` and rejects Python-specific object tags.

Validation:

- 71 focused YAML, configuration, stream, and file-I/O tests passed
- Original `StreamWriter` to `FileReader` reproduction passed
- Python compilation passed
- `git diff --check` passed

### Related

Fixes #1282
Follow-up: #1284

### DIFF
--- a/python/pyrogue/_HelperFunctions.py
+++ b/python/pyrogue/_HelperFunctions.py
@@ -336,22 +336,21 @@ def dataToYaml(data: Any) -> str:
         pass
 
     def _var_representer(dumper: yaml.Dumper, data: pr.VariableValue) -> Any:
-        """Represent ``VariableValue`` using display formatting and YAML typing."""
-        if isinstance(data.value, bool):
+        """Represent floats natively and other values by display text."""
+        if data.valueDisp is None:
+            return dumper.represent_scalar('tag:yaml.org,2002:null',u'null')
+        elif isinstance(data.value, bool):
             enc = 'tag:yaml.org,2002:bool'
         elif data.enum is not None:
             enc = 'tag:yaml.org,2002:str'
         elif isinstance(data.value, int):
             enc = 'tag:yaml.org,2002:int'
         elif isinstance(data.value, float):
-            enc = 'tag:yaml.org,2002:float'
+            return dumper.represent_data(data.value)
         else:
             enc = 'tag:yaml.org,2002:str'
 
-        if data.valueDisp is None:
-            return dumper.represent_scalar('tag:yaml.org,2002:null',u'null')
-        else:
-            return dumper.represent_scalar(enc, data.valueDisp)
+        return dumper.represent_scalar(enc, data.valueDisp)
 
     def _dict_representer(dumper: yaml.Dumper, data: odict) -> Any:
         """Represent ``OrderedDict`` values while preserving key order."""
--- a/python/pyrogue/_HelperFunctions.py
+++ b/python/pyrogue/_HelperFunctions.py
@@ -344,9 +344,11 @@ def dataToYaml(data: Any) -> str:
         elif data.enum is not None:
             enc = 'tag:yaml.org,2002:str'
         elif isinstance(data.value, int):
+            # Preserve readable integer formats such as hexadecimal in saved
+            # configs; unsafe display formats need a separate compatibility policy.
             enc = 'tag:yaml.org,2002:int'
         elif isinstance(data.value, float):
-            return dumper.represent_data(data.value)
+            return dumper.represent_float(float(data.value))
         else:
             enc = 'tag:yaml.org,2002:str'
 
--- a/tests/fileio/test_core_fileio.py
+++ b/tests/fileio/test_core_fileio.py
@@ -10,6 +10,7 @@
 
 import struct
 
+import numpy as np
 import pyrogue as pr
 import pyrogue.utilities.fileio
 import pytest
@@ -119,13 +120,21 @@ def test_file_reader_decodes_float_with_grouped_display_format(tmp_path):
         value=256939.708,
         disp="{:,.3f}",
     ))
+    writer.add(pr.LocalVariable(
+        name="NumpyBandwidth",
+        mode="RO",
+        value=np.float64(256939.708),
+        disp="{:,.3f}",
+    ))
     root.add(writer)
 
     with root:
         assert writer.Bandwidth.getDisp(read=False) == "256,939.708"
+        assert writer.NumpyBandwidth.getDisp(read=False) == "256,939.708"
         config_payload = root.getYaml(readFirst=False).encode("utf-8")
 
     assert b"256,939.708" not in config_payload
+    assert b"!!python/object" not in config_payload
 
     data_path = tmp_path / "grouped-float-config.dat"
     _write_record(data_path, 7, config_payload)
@@ -133,6 +142,7 @@ def test_file_reader_decodes_float_with_grouped_display_format(tmp_path):
     reader = pyrogue.utilities.fileio.FileReader(files=str(data_path), configChan=7)
     assert list(reader.records()) == []
     assert reader.configValue("root.DataWriter.Bandwidth") == 256939.708
+    assert reader.configValue("root.DataWriter.NumpyBandwidth") == 256939.708
 
 
 def test_file_reader_missing_path_raises(tmp_path):
--- a/tests/fileio/test_core_fileio.py
+++ b/tests/fileio/test_core_fileio.py
@@ -110,6 +110,31 @@ def test_file_reader_reads_records_and_config(tmp_path):
     assert reader.configValue("root.Child.Value") == 9
 
 
+def test_file_reader_decodes_float_with_grouped_display_format(tmp_path):
+    root = pr.Root(name="root", pollEn=False, initRead=False)
+    writer = pr.Device(name="DataWriter")
+    writer.add(pr.LocalVariable(
+        name="Bandwidth",
+        mode="RO",
+        value=256939.708,
+        disp="{:,.3f}",
+    ))
+    root.add(writer)
+
+    with root:
+        assert writer.Bandwidth.getDisp(read=False) == "256,939.708"
+        config_payload = root.getYaml(readFirst=False).encode("utf-8")
+
+    assert b"256,939.708" not in config_payload
+
+    data_path = tmp_path / "grouped-float-config.dat"
+    _write_record(data_path, 7, config_payload)
+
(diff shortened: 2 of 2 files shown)


python/pyrogue/_HelperFunctions.py
```diff
@@ -336,22 +336,21 @@ class PyrogueDumper(yaml.Dumper):
...
+            return dumper.represent_scalar('tag:yaml.org,2002:null',u'null')
+        elif isinstance(data.value, bool):
             enc = 'tag:yaml.org,2002:bool'
         elif data.enum is not None:
             enc = 'tag:yaml.org,2002:str'
         elif isinstance(data.value, int):
             enc = 'tag:yaml.org,2002:int'
         elif isinstance(data.value, float):
-            enc = 'tag:yaml.org,2002:float'
+            return dumper.represent_data(data.value)
```
ruck314: `np.float64` subclasses `float`, so numpy scalars reach this branch, but `represent_data` dispatches on the exact type and falls through the mro loop to `Representer.represent_object`. Verified with PyYAML 6.0.3 / numpy 2.2.6, a `np.float64(1.5)` value serializes as:

```yaml
Bandwidth: !!python/object/apply:numpy._core.multiarray.scalar
- !!python/object/apply:numpy.dtype
  args: [f8, false, true]
  state: !!python/tuple [3, <, null, null, null, -1, -1, 0]
- !!binary |
  AAAAAAAA+D8=
```

237 bytes instead of 3, rejected by `yaml.safe_load`, and the embedded module path is `numpy._core` on numpy 2.x vs `numpy.core` on 1.x, so configs stop round-tripping across a numpy major bump.

Not a regression for rogue core (every float model returns a Python float via `struct.unpack`, and `Fixed`/`UFixed` go through the C++ `rim.Fixed` path), but downstream `LinkVariable` / `LocalVariable` getters that do numpy math are common, and they would silently write corrupt configs.

```suggestion
            return dumper.represent_float(float(data.value))
```

Verified across the edge cases: `256939.708` to `256939.708`, `np.float64(1.5)` to `1.5`, `nan`/`inf` to `.nan`/`.inf`, `1e-12` to `1.0e-12`, all exact on reload. It also keeps the explicit `tag:yaml.org,2002:float` tag rather than relying on implicit resolution.

python/pyrogue/_HelperFunctions.py
```diff
@@ -336,22 +336,21 @@ class PyrogueDumper(yaml.Dumper):
...
-        """Represent ``VariableValue`` using display formatting and YAML typing."""
-        if isinstance(data.value, bool):
+        """Represent floats natively and other values by display text."""
+        if data.valueDisp is None:
+            return dumper.represent_scalar('tag:yaml.org,2002:null',u'null')
+        elif isinstance(data.value, bool):
             enc = 'tag:yaml.org,2002:bool'
         elif data.enum is not None:
             enc = 'tag:yaml.org,2002:str'
         elif isinstance(data.value, int):
```
ruck314, line 346: The same bug is still live on this branch: `disp='{:,d}'` produces `!!int '1,234,567'`, which fails with the identical `ValueError` (confirmed locally). #1282's suggested fix says "numeric variables", and this PR covers only float.

There is a real trade-off against changing it, since `disp='{:#x}'` currently keeps register masks readable in hand-edited configs and a native int would turn those decimal. Whichever way you go, please add a short comment stating that int deliberately keeps the disp text and why, so the next person does not file the same issue against `{:,d}`.

tests/fileio/test_core_fileio.py
```diff
@@ -110,6 +110,31 @@ def test_file_reader_reads_records_and_config(tmp_path):
...
+        config_payload = root.getYaml(readFirst=False).encode("utf-8")
+
+    assert b"256,939.708" not in config_payload
+
+    data_path = tmp_path / "grouped-float-config.dat"
+    _write_record(data_path, 7, config_payload)
+
+    reader = pyrogue.utilities.fileio.FileReader(files=str(data_path), configChan=7)
+    assert list(reader.records()) == []
+    assert reader.configValue("root.DataWriter.Bandwidth") == 256939.708
```
ruck314, line 144: Please cover `np.float64` here too, since that is the input that breaks the new float branch (see the comment on `_HelperFunctions.py`). A second `LocalVariable` with `value=np.float64(256939.708)`, asserting the payload contains no `!!python/object` tag, would pin it.

Optional: this exercises a synthetic `LocalVariable` rather than the `StreamWriter` / `DataWriter.Bandwidth` path from #1282. Fine and cheaper as a unit test, but a case going through the real variable would keep the regression pinned to the reported path if the `_DataWriter.py:131` disp ever changes.

ruck314 (changes requested):
Approach is right: fixing the serializer rather than the `Bandwidth` disp string is the correct call, since `disp` is a display concern and should never gate round-trippability. Hoisting the `valueDisp is None` check to the top of the chain is behavior-preserving and reads better than the old trailing check. bool/enum/int/str branches are untouched, `getDisp()` still returns `256,939.708` for the UI, and ndarray/list variables are unaffected because `_getDict` always builds `VariableValue` with `index=-1`, so `value` is the container and falls to the str branch.

One blocking item inline: `represent_data()` is the wrong primitive and corrupts numpy float scalars. It is a one-line fix.

Two smaller items inline: the identical bug is still live for int disp formats, and the new test should cover `np.float64`.

One thing for the release notes, not a change request: every float variable's serialized value changes. `disp='{:1.3f}'` on `0.123456789` used to write `0.123` and reload the rounded value; it now writes and reloads full precision. That is more correct and is exactly the class of bug in #1282, but the saved-config text churns for every downstream repo doing golden-file comparisons.

On the "flake8 was not available" note: CI covers it. `scripts/run_linters.sh` (.github/workflows/rogue_ci.yml:53) runs compileall + flake8 + cpplint.

python/pyrogue/_HelperFunctions.py
bengineerd: Fixed in c5908f239. The float branch now uses `dumper.represent_float(float(data.value))`, which normalizes `np.float64` to a portable Python float while retaining the explicit YAML float representation. The regression test now covers the NumPy scalar path and verifies `FileReader` round-tripping.

python/pyrogue/_HelperFunctions.py
bengineerd: Kept integer serialization behavior unchanged in this PR to preserve readable formats such as hexadecimal register values. Added an inline comment documenting that compatibility choice, and opened #1284 to design safe fallback behavior for non-YAML formats such as comma-grouped decimal integers without regressing hex output.

tests/fileio/test_core_fileio.py
bengineerd: Added `np.float64(256939.708)` coverage in c5908f239. The test now asserts that no `!!python/object` tag is emitted and verifies both Python and NumPy float values through the `FileReader` config-channel path. I kept the synthetic variable because it directly pins the serializer contract; the PR validation also includes the original real `StreamWriter`/`DataWriter.Bandwidth` reproduction.

ruck314 (approved):

PR #1286, opened 2026-08-19
Add a deterministic-counter performance gate that compares each pull request to its own merge base

ruck314:
### Description

This change adds a performance gate that measures every pull request against its own merge base, built and measured together in the same job, and blocks a pull request only on deterministic counts of work performed: buffer copies, heap allocations, frame and buffer constructions, mutex acquisitions, and boundary crossings between Python and native code. Wall clock throughput numbers keep being collected and published exactly as before and never gate a pull request. When a gating counter cannot be measured cleanly, the gate reports a separate, non-blocking inconclusive verdict instead of guessing, so a red check always means a real, deterministic difference from the merge base.

### Details

**What this pull request's own check reported, worth reading before looking at the check itself.** The gate workflow already lives in this branch's tree, so it ran on this pull request like any other: GitHub Actions run 32297377577 completed in 14 minutes 13 seconds with every step succeeding, including the paired measurement. Its merge base leg built the base branch, which carries no gate baseline file at all, so the gate resolved through its own distinct structural condition for an absent trusted baseline, which short-circuits ahead of any per-cell evaluation, leaving zero gating cells evaluated. The result was a non-blocking inconclusive verdict, the gate's third state working as designed, not the feature failing on arrival, and the check therefore reported success rather than a failure. Separately, a fork pull request runs with a cold compiler cache by construction, and the two recorded cold-cache samples measured during this work were 624 seconds and 758 seconds against the gate's own 600 second budget, so this pull request's own run is likely to exceed that budget even though the paired measurement step by itself has generally finished well inside it.

**What was measured**, condensed from the completion record this branch carries at `docs/plans/perf-ci-hardening/COMPLETED.md`, which every figure below matches exactly. The tier registry binds 22 metrics to the gating tier because each was proven bit identical across the three processor models this campaign observed, two AMD EPYC models and one Intel Xeon model; the committed baseline resolves that into 128 gating cells across 16 benchmarks. Ten real pull-request runs against changes that could not legitimately regress anything produced zero failing gating cells. Of five deliberately seeded regressions, four were detected and attributed on real hosted records; the fifth, a disabled-frame-batching patch, produced no usable sample across eleven hosted dispatches and is closed by an explicit, attributed override rather than a detection. The paired same-job comparison's own measured cost sits at a median of 350.77 seconds against the 600 second budget, computed over the 73-record population the completion record cites.

**What is not yet closed**, carried forward from the completion record rather than composed again:

- The disabled-frame-batching regression's second lock has never been isolated on any population across eleven hosted dispatches; it is closed for now by an explicit override carrying no measured magnitude, not by a detection.
- The small per-transaction timing regression this project seeded is invisible to the gate by design, because the deterministic tier deliberately never gates on wall clock or processor time.
- A calibration-normalized second-tier gating comparison has never been exercised on any population.
- No shipped benchmark measures a normalized processor-time-per-unit-of-work quantity end to end.
- Gate behavior under a real, nonzero cloud-provider processor steal delta has never been observed on the runner fleet this project measured, so it remains untested.
- Three probe defects from the runner capability work remain open pending a fresh probe: a zero-counted instrument graded usable rather than demoted, and a repeatability gate that has only ever been exercised at one repeat.
- The trailer-parser-unavailable fail-closed path of the escape hatch has only ever been proven by unit coverage, never fired on a real hosted run.
- The trusted-baseline residual risk stays open by design: a baseline change that reaches the base branch through ordinary review is trusted once it lands, and the one disclosure signal that exists does not force a reviewer to read it.
- Branch protection on the real target repository has not been applied yet; see the checklist item below.
- A closing account of this work must read its own code review, not only its own plans, a practice this record commits to going forward rather than treating as a one-time fix.
- Eighteen references to this project's own private planning identifiers survive inside two generator-produced files, twelve in the rendered campaign report and six in its JSON sidecar, because removing them means editing and regenerating both, and regeneration was deliberately excluded here since the byte-stability check tying those exact bytes to the already-committed files has twice caught a real generator defect in this project's history. The same twelve references also live inside the generator's own source for the identical reason, one further reference lives inside a workflow file's own step-output text rather than a comment and was left untouched because this project's rule for that pre-existing file forbids changing that kind of text, and one separate reference in an interface test module predates this work entirely and was left untouched on purpose.

**What ships in this pull request and what does not.** Every generated report and its versioned sidecar, the probe population, the seeded-regression patches, the retained `gate-runs` directory (including one full paired-run record kept as the offline re-score worked example), the decisions record, the module audit, the recorded smoke test, and the completion record itself all ship in this pull request's tree. The full paired-run population and the two harness measurement populations do not ship, because neither is read by path anywhere under `tests/`, `scripts/`, or the workflows, and both are too large for a pull request to carry and still be read line by line. Both remain reachable through the annotated tag `perf-ci-hardening-evidence-v1` already published on this fork, whose name and resolved commit are recorded once in `docs/plans/perf-ci-hardening/gate-runs/evidence-tag.json`. That same tag also carries this effort's own intermediate development history, the individual incremental commits these eight commits were cut from; if this fork ever disappears, that history and those two raw populations go with it, along with the ability to regenerate a report from its raw input or recheck the paired-mechanism cost figures against the population they were computed from. Every figure backed by a retained per-run artifact under `gate-runs/` stays checkable from this pull request's own tree alone.

**One action a maintainer must still take, and no commit in this pull request can perform it:**

- [ ] Add `Perf Gate` (the gate job's own frozen name) to the base branch's existing required status checks, once this pull request's own gate job has completed successfully against the real target repository at least once. Until that setting is applied, the gate exits non-zero on a failing verdict, but nothing on this repository's own side enforces it, so it stays advisory in practice.

**A suggested reading order for a reviewer.** These eight commits were cut along real subsystem seams for review and are not the shape the work actually happened in; the full incremental history lives on the fork tag named above.

1. `tests/perf/README.md` first, the shortest complete statement of what the gate does and what a red check means.
2. `scripts/perf_gate_evaluator.py`, the single function that turns a two-leg measurement record into a verdict.
3. `scripts/perf_gate_ab_runner.py`, the paired same-job runner that produces that record.
4. `.github/workflows/perf_gate.yml`, the workflow that wires the runner and the evaluator into a pull request.
5. `scripts/perf_tier_registry.py` and `tests/perf/_perf_harness.py`, the metric registry and the measurement harness underneath both.
6. The test modules under `tests/perf/` and `tests/utilities/`.

The generated reports and the retained evidence under `docs/plans/perf-ci-hardening/` are reference material to sample for a specific figure, not files meant to be read start to finish.


ruck314:
## Could this gate exist without modifying the library?

Recording this for the record, since the production-code footprint is the main reason to be uneasy about this pull request and the answer is not a simple yes or no.

### What the footprint actually is

Of the 161 files here, the library itself accounts for 20: 17 under `src/` and 3 under `include/`, totalling 518 insertions and 3 deletions. Everything else is tests, scripts, workflows, and evidence. There are no Python library changes. The 3 deletions mean this is almost entirely additive.

The substantive parts are a new `rogue::perf` translation unit holding 13 process-wide counters, and increment sites in the stream copy helpers, `Pool`, `FrameLock`, `TransactionLock`, `GilRelease`, `ScopedGil`, the memory `Master`/`Transaction`, and the TCP and UDP paths.

### The case that it could not be done externally

The gate blocks only on counters that were bit-identical across 24 clean runs and 3 processor models. That determinism bar is what makes a red check mean something, and it is also what rules out the non-invasive options. This was measured rather than assumed, in `docs/plans/perf-ci-hardening/CAPABILITY-MATRIX.md`, over 20 probe runs per instrument with a 5x overhead disqualification threshold:

| Instrument | Median overhead | Outcome against real modules |
|---|---|---|
| Call-graph profiler (callgrind) | 190.5x | Never wired in; incompatible with any sane job timeout |
| Syscall tracer (strace) | 9.5x | Disqualified on overhead before reaching real modules |
| Preloaded allocation interposer | 1.07x | Counts unstable: 26 of 40 runs not bit-identical |
| perf software events | 0.61x | Insufficient samples across all 40 runs |

The two instruments cheap enough to run were not deterministic enough to gate on. The one that would have been deterministic costs 190x. That is the core finding, and it is the strongest argument that in-library counters were the only route to this particular gate.

There is also a structural obstacle independent of cost. The hottest quantities live in code that has no symbol to attach to from outside. `toFrame`, `fromFrame`, and `copyFrame` are `static inline` in `FrameIterator.h`, so the `memcpy` calls are compiled into every caller as an intrinsic rather than reached through the PLT. `FrameLock` and `TransactionLock` are RAII types that inline away. An `LD_PRELOAD` shim, a uprobe, or an eBPF attachment can only hook what survives as a callable symbol, and these do not. Separately, an allocation interposer sees `malloc` but cannot distinguish a `Pool` buffer allocation from any other allocation in the process, and nothing outside the library can observe a GIL crossing as a distinct event.

So: a gate on *these* metrics, at *this* determinism bar, is not reachable from outside the library.

### The case that the footprint should have been much smaller

That argument justifies the counters existing. It does not justify the specific form they took, and there are three concrete criticisms that stand.

**They are always on.** Every counter is unconditional. `Directives.h` and `GilRelease.cpp` already establish the compile-time-conditional pattern in this codebase through `NO_PYTHON`, so gating the whole facility behind something like `ROGUE_PERF_COUNTERS`, defaulting to off, would have followed existing house style and left shipped builds carrying nothing. The gate would still work, because CI controls its own build flags. This is the single change that would have addressed most of the objection, and it was not made.

**The counters are not cache-line padded.** All 13 are declared consecutively as 8-byte atomics, so the first eight share one 64-byte line. `gBufferCopyBytes` and `gBufferCopyCount` are adjacent and both incremented on every `memcpy` chunk, which means two read-modify-write operations on the same line per chunk, and on x86 a relaxed `fetch_add` is still a locked instruction rather than a free one. In a multi-threaded streaming workload every thread doing frame copies contends on that one line. For a high-rate data acquisition library this is the wrong default, and it is trivially fixable with padding or per-thread accumulation.

**`Pool` gained three atomic members.** Adding them to the class in a public header changes `sizeof(Pool)`, which breaks ABI for anything compiled against the previous header. A side table keyed by pool instance, or moving the state behind the existing indirection, would have kept the counters without that consequence.

**There was also a middle path.** Because the gate compares a pull request against its own merge base and applies identical instrumentation to both legs, the comparison stays valid if the instrumentation is never merged at all and is instead applied as a patch in CI. The counters measure semantic events rather than speed, so an instrumented build is a legitimate place to count them. The cost is maintaining a patch that must apply to arbitrary older merge-base trees, against files that change rarely. That path keeps the library untouched and was not taken.

### What a future attempt should do differently

Keep the paired same-job design and the deterministic-counter idea; both are validated here. Change the delivery: gate the facility off at compile time, pad the counters or make them thread-local with aggregation at read time, keep `Pool`'s layout stable, and decide deliberately whether the instrumentation belongs in the library at all or in a CI-applied patch.

### If this is closed

The evidence is worth more than the code. The full measurement populations, including the capability matrix behind the table above, are reachable through the annotated tag `perf-ci-hardening-evidence-v1` on the fork, recorded in `docs/plans/perf-ci-hardening/gate-runs/evidence-tag.json`. The instrument overheads and the determinism results are the reusable part: they say what does not work and why, which is most of the cost of a second attempt.

One correction to the record while closing: this pull request's own check reports a non-blocking inconclusive because the base branch carries no baseline file, which is the gate's designed third state and not a failure.

ruck314:
Closing this rather than pursuing it. The reason is the library footprint: the counters are always on, they are not cache-line padded, and they change `Pool`'s layout. The preceding comment records what was measured about whether the gate could have avoided touching the library at all, and what a future attempt should do differently. Leaving that here as the record.

PR #1289, opened 2026-09-09
Fix float display value parsing

bengineerd:
### Description
Ensure float variables coerce display-parsed values to Python float, including whole-number input, and reject boolean display values instead of accepting them as numeric floats.

### Details
Added regression coverage for whole-number and boolean display input.

Validation:
- `tests/core/test_model_variables.py`: 8 passed
- Python compile checks passed for both changed files
- `git diff --check` passed

### DIFF
--- a/python/pyrogue/_Variable.py
+++ b/python/pyrogue/_Variable.py
@@ -877,7 +877,7 @@ class BaseVariable(pr.Node):
         """
         try:
             if not isinstance(sValue,str):
-                return sValue
+                value = sValue
             elif self.nativeType is np.ndarray:
                 return np.array(ast.literal_eval(sValue),self._ndType)
             elif self.disp == 'enum':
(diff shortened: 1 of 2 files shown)


python/pyrogue/_Variable.py
```diff
@@ -877,15 +877,22 @@ def parseDisp(self, sValue: str) -> object:
...
                 return np.array(ast.literal_eval(sValue),self._ndType)
             elif self.disp == 'enum':
                 return self.revEnum[sValue]
             elif self.nativeType is str:
                 return sValue
             else:
-                return ast.literal_eval(sValue)
+                value = ast.literal_eval(sValue)
+
+            if self.nativeType is float:
```
ruck314: Ryan is recommending this:
```
        try:
            if not isinstance(sValue,str):
                value = sValue
            elif self.nativeType is np.ndarray:
                return np.array(ast.literal_eval(sValue),self._ndType)
            elif self.disp == 'enum':
                return self.revEnum[sValue]
            elif self.nativeType is str:
                return sValue
            else:
                value = self.nativeType(ast.literal_eval(sValue))

            return value
```

ruck314 (approved):

Issue #1294, opened 2026-09-18
VirtualClient monitor thread stays alive: any script holding a client hangs at exit


tristpinsm:
## Summary

`VirtualClient._monThread` is constructed without `daemon=True` on `main` and
`pre-release` (`python/pyrogue/interfaces/_Virtual.py:513`), so any process that
builds a `VirtualClient` and does not explicitly call `stop()` **never exits**.

This is #1238, which PR #1239 fixed on 2026-05-14. That fix was reverted on
2026-05-31 by PR #1252, one of a batch of five reverts (#1248–#1252), and the
regression shipped in **v6.15.0**.

This has surfaced again as I am building a new client interface for the `pysmurf`
refactor. I've worked around it there and similarly in `pysmurf` for the rogue 6
upgrade, but it would be great if we could find a solution within `rogue`.

The following was generated by Claude based on tests in different configurations.
One interesting finding from those measurements is that making the thread a daemon
is not enough to completely prevent this hang. It seems like an explicit tear-down
routine may be needed to exit cleanly.

## Affected versions

Line history:

```
016a8ad10 Revert changes from PR#1239        -> daemon=True removed
4aab1d63a fix(virtualclient): daemon ... #1238 -> daemon=True added
c3b71c8cb Convert timers to threads          -> thread introduced, non-daemon
```

## Reproducer — no hardware, no pysmurf

Server:

```python
# server.py
import sys, time
import pyrogue as pr
import pyrogue.interfaces

root = pr.Root(name='AMCc', description='minimal server')
root.add(pr.LocalVariable(name='Dummy', value=0))
root.addInterface(pr.interfaces.ZmqServer(root=root, addr='localhost', port=int(sys.argv[1])))
root.start()
print('SERVER UP', flush=True)
try:
    while True:
        time.sleep(1)
finally:
    root.stop()
```

Client — note it does nothing unusual; it connects and falls off the end:

```python
# client.py
import sys
from pyrogue.interfaces import VirtualClient

c = VirtualClient('localhost', int(sys.argv[1]))
print('linked', c.linked, 'daemon', c._monThread.daemon, flush=True)
```

```console
$ python3 server.py 18760 &          # binds 18760 and 18761
$ timeout 15 python3 client.py 18760
Connected to AMCc at localhost:18760
linked True daemon False
$ echo $?
124                                   # 124 == timeout killed it; it never exits
```

5 of 5 trials hang. The interpreter is finished — `main thread done` has printed —
[...]

Issue #1238, opened 2026-05-13
ZMQ client monitor thread is not exiting in interactive environment


tristpinsm:
The monitor thread of the python `VirtualClient` is not exiting when the client is destroyed. For instance, a pysmurf client object which holds a `VirtualClient` instance in an `ipython` session hangs when the session is exited.

```
In [2]: S.setup()
[ 2026-05-13 22:30:12 ]  Setting up...
[ 2026-05-13 22:30:12 ]  Toggling DACs
[ 2026-05-13 22:30:12 ]  Waiting 20.00 seconds after...
WARNING:pyrogue.VirtualClient.VirtualClient:I have not heard from AMCc in 10.0 seconds. It may be busy, continuing to wait...
WARNING:pyrogue.VirtualClient.VirtualClient:I have finally heard from AMCc. All is good!
[ 2026-05-13 22:30:32 ]  Done waiting.
[ 2026-05-13 22:31:07 ]  System configuration finished after 34 seconds. The final state was True.
[ 2026-05-13 22:31:34 ]  Turning off tones
[ 2026-05-13 22:31:34 ]  Turning off flux ramp
[ 2026-05-13 22:31:34 ]  Turning off all TES biases
[ 2026-05-13 22:31:35 ]  Percent full scale = 49.500%
[ 2026-05-13 22:31:35 ]  Attempts to communicate with a cryocard failed!  Will assume no cryostat card is connected and skip cryocard setup steps.
[ 2026-05-13 22:31:35 ]  Configuring the system to take timing from ext_ref
[ 2026-05-13 22:31:35 ]  System already configured for ext_ref timing, doing nothing.
[ 2026-05-13 22:31:35 ]  Done with setup.
Out[2]: True

In [3]: quit()

[hang forever]
```

This was worked-around in `pysmurf` by calling the `stop` method on `VirtualClient` to disable the monitor thread. Recent versions of `rogue` have made that `stop` call also close the ZMQ connection. Either we can find a way to make the monitor thread exit, or `pysmurf` can explicitly disable it without calling `stop`.

tristpinsm:
In this case, the software was built with

- `rogue` C code compiled at `1a04b1e1fd0df83c41cd45bc25`, but python is a checkout of branch `issues/1234`
- `pysmurf` on `v11.0.3`
- `cryo-det` `v2.4.0`

tristpinsm:
here's a new workaround on the `pysmurf` side: slaclab/pysmurf#991

[...]

PR #1295, opened 2026-09-18
fix(virtualclient): tear down monitor thread at interpreter exit

ruck314:
### Description

`VirtualClient` now tears itself down at interpreter exit, so a script that constructs a client and never calls `stop()` exits instead of hanging. Restores `daemon=True` on the monitor thread and adds a process-wide shutdown hook that runs the full `stop()`, releasing the C++ `ZmqClient` before the interpreter finalizes. Calling `stop()` explicitly is still supported for callers that want the connection released earlier.

### Details

- Regression of #1238: PR #1239's fix was reverted by #1252 and shipped broken in v6.15.0.
- `daemon=True` alone is not enough. It trades the guaranteed hang for an intermittent `SIGABRT` when the C++ client is still live at shutdown; only the full `stop()` measured clean (30/30 locally, was 124-timeout 5/5 before).
- `atexit` cannot fix this: `threading._shutdown()` runs its joins first, so the hook uses `threading._register_atexit` with an `atexit` fallback. Rogue holds that private-API dependency so downstream callers do not have to.
- The hook sweeps `ClientCache` rather than registering per instance, since `_register_atexit` has no unregister. `stopMonitor()` is deliberately not reintroduced.
- New `test_virtual_client_exit_hook.py` verified failing against the unfixed tree. Local: linters exit 0, `tests/interfaces` 98 passed, `ctest -L cpp` 18/18.

### Related

Closes #1294. Prior history: #1238, #1239, #1252.


ruck314:
I relooked at https://github.com/slaclab/rogue/pull/1252 and this is a feature that Ryan did want. closing this PR

